PR #19, opened 2021-04-18
Discord support

TurtIeSocks:
- Passport.js
- Discord strategy
- Discord oauth methods (thanks @versx / MapJS )
- MySQL Sessions
- Dynamic database selection
- First migration for users table, more to come
- Initial login page to show to unauthorized users. More to come.
- Discord Perms now set the default filters & load the ui elements available to them from the backend
- Updated config example/default
- Added some defaults for db stuff
- Dynamic importing for each category of map items
- New function to capitalize/split camel case words
- Combined user/settings into one request
- Updated yarn scripts for properly starting dev/prod environments


server/src/services/session-store.js
```diff
@@ -0,0 +1,58 @@
...
+  return results.length < config.maxSessions
+}
+
+const clearOtherSessions = async (userId, currentSessionId) => {
+  const results = await Session.query()
+    .where(raw(`json_extract(data, '$.passport.user.id') = '${userId}'`))
+    .andWhere('session_id', '!=', currentSessionId)
+    .delete()
+  console.log('[Session] Clear Result:', results)
+}
```
versx, line 52: This is some sexy code...

server/src/services/ui.js
```diff
@@ -0,0 +1,36 @@
...
+      }
+    }
+    if (key === 'gyms') {
+      filterItems.gyms = {}
+      const secondary = ['gyms', 'raids']
+
+      for (let i = 0; i < secondary.length; i += 1) {
+        if (filters[secondary[i]]) filterItems.gyms[secondary[i]] = true
+      }
+    }
```
versx, line 31: What do you think about using a switch instead of the multiple if statements?

TurtIeSocks: Good call, I always forget about switch!

PR #71, opened 2021-06-08
Mad support

TurtIeSocks:
- Adds Full MAD Support 
- PVP Sliders are disabled for MAD, for now
- dbSelection function to proper check what each kind of db is and what it's being used for
- Moved a lot of queries to the respective models and out of the `schema.js` file
- Add en-gb language file
- Fix level filtering
- Existing RDM users should make sure to add a `type` key to their schemas added in their config


server/src/configs/default.json
```diff
@@ -23,7 +23,8 @@
       "quests": false,
       "raids": false,
       "nests": false
-    }
+    },
+    "weatherHistoryDays": 3
```
versx, line 27: :feelslapras: Weather History _Days_????

TurtIeSocks: 😅 couldn't think of a better key name. Thank you for the reminder though to add to the PR notes!

versx (approved):

server/src/configs/default.json
versx: 🤔 Hmm good point, not really sure what you would call it, maybe  `weatherCellScanTimeLimit` possibly or sometime similar? Just throwing random names out here.

TurtIeSocks: Probably makes sense to have `limit` in there somewhere at the very least.

PR #78, opened 2021-06-09
Adds Optional Discord Array

TurtIeSocks:
- Can add specific categories to this array that allows users to view those categories without having to login to Discord, even when Discord is enabled.
- `false` on each perm now makes it not viewable even if there are perms in the role array
- Logout button can now be a login button if the user is not logged in
- Logout/Login button does not appear if discord is disabled

### DIFF
--- a/server/src/configs/config.example.json
+++ b/server/src/configs/config.example.json
@@ -68,6 +68,7 @@
       }
     }  
   },
+  "alwaysEnabledPerms": [],
   "discord": {
     "enabled": false,
     "logChannelId": "",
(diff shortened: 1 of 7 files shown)


lenisko:
> Can add specific roles to this array that allows users to view those categories without having to login to Discord, even when Discord is enabled.

Wat? 😂

TurtIeSocks:
> > Can add specific roles to this array that allows users to view those categories without having to login to Discord, even when Discord is enabled.
> 
> Wat? 😂

Rip. Fixed 😰

PR #85, opened 2021-06-11
SQL Area Restrictions

TurtIeSocks:
Adds SQL Area Restrictions for:
- Devices
- Gyms
- Nests
- Pokemon
- Pokestops
- S2 Scan Cells
- Spawnpoints
- Unlimited number of rules can be set in the config

Partially authored by @Len for MapJS.

Co-Authored-By: Jakub <10072920+lenisko@users.noreply.github.com>

### DIFF
--- a/server/src/configs/config.example.json
+++ b/server/src/configs/config.example.json
@@ -81,6 +81,16 @@
     "allowedGuilds": [],
     "blockedGuilds": [],
     "allowedUsers": [],
+    "areaRestrictions": {
+      "rule1": {
+          "roles": [],
+          "areas": []
+      },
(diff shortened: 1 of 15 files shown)


lenisko (approved):
Had one freeze/crash which I couldn't track due to the lack of debug logs 🤔but been working just fine after rebuild.
Aside one commented line of code can't see anything "bad" in code. As it's beta I would push it for general use, so maybe we'll get more feedback. Proposed change in configuration looks good to me, object keys been unused before and in case we'll need some naming, we can just put it inside array objects.

Good job overall!

versx (approved):

PR #86, opened 2021-06-11
Add option to offset raid boss/egg/quest rewards on markers

Petap0w:
Adds advanced config options to allow offsetting the raid boss/egg on gym marker, and the quest reward on pokestop marker.
This gives option to replace the icon set and have markers nicely arranged.
Example with :
```
      "iconModifiers": {
        "gyms": {
          "offsetX": 0.2,
          "offsetY": 0.3
        },
        "pokestops": {
          "offsetX": 0.2,
          "offsetY": 0.1,
          "sizeMultiplier": 0.7
        }
      }
```
Capture d’écran 2021-06-11 à 10 19 40
Capture d’écran 2021-06-11 à 10 33 01

### DIFF
--- a/server/src/configs/default.json
+++ b/server/src/configs/default.json
@@ -95,6 +95,17 @@
         "lg": 2,
         "xl": 3
       }
+    },
+    "iconOffsets": {
+      "gyms": {
+        "offsetX": 0,
+        "offsetY": 0
+      },
+      "pokestops": {
+        "offsetX": 0,
+        "offsetY": 0,
+        "sizeMultiplier": 0
+      }
     }
   },
   "clientSideOptions": {
@@ -123,7 +134,7 @@
       "glow": [
         {"name": "Hundo", "perm": "iv", "num": 100, "value": "#ff1744", "op": "=" },
         {"name": "Top 3 Ranks", "perm": "pvp", "num": 3, "value": "#0000ff", "op": "<=" },
-        {"name": "Multiple", "perm": "pvp", "value": "#800080" }    
+        {"name": "Multiple", "perm": "pvp", "value": "#800080" }
       ]
     },
     "wayfarer": {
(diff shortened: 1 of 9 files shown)


server/src/configs/default.json
```diff
@@ -75,13 +75,18 @@
         "sm": 20,
         "md": 35,
         "lg": 50,
-        "xl": 65
+        "xl": 65,
+        "offsetX": 0,
+        "offsetY": 0
```
TurtIeSocks, line 80: We should add offsets to the specific icon packs instead of in the universal sizes section. This will be different for every icon set, which can be changed on the client side so even if you get these dialed in perfectly for whatever set *you* prefer, maybe your users prefer the other icons and will now not be aligned properly.

server/src/configs/default.json
```diff
@@ -75,13 +75,18 @@
...
       },
       "pokestops": {
         "sm": 15,
         "md": 25,
         "lg": 35,
-        "xl": 45
+        "xl": 45,
+        "offsetX": 0,
+        "offsetY": 0,
+        "sizeMultiplier": 0
```
TurtIeSocks, line 89: Same as above.

TurtIeSocks (changes requested):
Moving offsets to the icons section (where the URLs are) provides a lot more flexibility for users to choose what iconset they want and ensures that they are all properly aligned. It will be a little more work to get that data translated to the frontend but not too bad. I would recommend using a format like this to make sure the "category" keys on the frontend align with th esame keys:

```json
"icons": {
  "Default": {
    "path": "https://mygod.github.io/pokicons/v2",
    "offsets": {
      "pokestops": {
         "x": 0,
         "y": 0
      },
      "gyms": {
         "x": 0,
         "y": 0
      }
    }
  }
}
```

PR #88, opened 2021-06-12
Allow enable/disable of item in advanced filters by click on image

Petap0w:


TurtIeSocks (changes requested):
This works perfectly fine but one small nitpick. Since the onclick function is the same, I would prefer if we moved it to its own function, then called it on both the Icon and the Img, to reduce repeat code, see below:
Screen Shot 2021-06-12 at 12 02 45 PM

PR #92, opened 2021-06-14
Makes location accuracy circle overlay a config setting

Petap0w:
When sharing connection there is sometimes a large location accuracy circle overlay which makes the map content not really visible.


TurtIeSocks (changes requested):
I don't think we should remove it entirely as this seems like a smaller use case than those who don't share mobile connections. I would suggest, at most, we reduce the opacity. 
```js
  const locateOptions = {
    keepCurrentZoomLevel: true,
    circleStyle: { fillOpacity: 0.05 },
    strings: {
      title: t('useMyLocation'),
    },
    onActivate: () => { },
  }
```
Alternatively, put it in Settings.

lenisko:
Settings ++

lenisko:
I think that client side switch would make more sense. Sorry for not pointing it before... @TurtIeSocks ?

Petap0w:
when you don't have gps issues you won't even have this circle, and when you have gps problems you'll want it off to see the map (and accuracy is not really an important information anyways when using the map, you'll just want the screen to be centered "around" your location when clicking on location button so you don't have to scroll).

lenisko:
For me it happens quite often that I lose a location on the map (thank you Android) so it's nice to have. Making it client side setting users would be able to decide whatever they need this or not at the current time. For starter maybe @TurtIeSocks solution to make it more transparent?

server/src/configs/default.json
```diff
@@ -30,10 +30,11 @@
     "title": "ReactMap",
     "headerTitle": "ReactMap",
     "nestHemisphere": "north",
-    "enableFeedbackLink": false,
+    "enableFeedback": false,
```
TurtIeSocks, line 33: I would remove this change from this PR since it's coming in your local auth PR anyway, just to reduce conflicts. After that, I can merge this in.

PR #93, opened 2021-06-14
Adds title description when mouse over the menu / location / zoom buttons

Petap0w:
Adds title description when mouse over the menu / location / zoom buttons

### DIFF
--- a/src/components/ConfigSettings.jsx
+++ b/src/components/ConfigSettings.jsx
@@ -128,6 +128,8 @@ export default function ConfigSettings({
           <Map
             serverSettings={serverSettings}
             params={match.params}
+            zoomControl={false}
+            locateControl={false}
           />
         )}
       </MapContainer>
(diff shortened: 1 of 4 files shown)


src/components/ConfigSettings.jsx
```diff
@@ -128,6 +128,8 @@ export default function ConfigSettings({
           <Map
             serverSettings={serverSettings}
             params={match.params}
+            zoomControl={false}
```
TurtIeSocks, line 131: I guess I'm slightly confused why you added zoomControl and locateControl here? <Map /> component is one of ReactMap's component, not a React Leaflet Component, you can see on line 124 in the <MapContainer /> component, ZoomControl is already being disabled.

TurtIeSocks (changes requested):
I tried adding `locateControl={false}` to <MapContainer /> and it didn't seem to make a difference. I appreciate the titles/translations you've added but I'm not seeing what the other parts of the PR add. I still see the invisible locateControl div in the Chrome dev panel.

PR #95, opened 2021-06-15
Tutorials & Profile Page

TurtIeSocks:
- Adds user profile page that shows current permissions
- Adds step by step tutorial that covers the basics
- Fixes some Pokemon filtering bugs
- More consistent coloring/behavior for global filters
- IV Global filter now shows nothing at 0-100, like all of the others
- 100% IV switch


src/components/layout/dialogs/tutorial/Welcome.jsx
```diff
@@ -0,0 +1,77 @@
...
+export default function TutWelcome({ setUserProfile }) {
+  const { t } = useTranslation()
+  const { loggedIn, perms } = useStatic(state => state.auth)
+  const { map: { excludeList } } = useStatic(state => state.config)
+
+  const getPerms = () => {
+    let have = 0
+    let total = 0
+    Object.keys(perms).forEach(perm => {
+      if (!excludeList.includes(perm)) {
```
lenisko, line 23:
```
TypeError: excludeList is undefined
    getPerms https://.../main.877f77aa93add619f148.js:1625
    getPerms https://.../main.877f77aa93add619f148.js:1624
    TutWelcome https://.../main.877f77aa93add619f148.js:1662
```

lenisko (changes requested):
Gray page after fresh login.

lenisko (changes requested):
I like this idea modal is working fine! Two things though 😨 
Many users tend to use few devices to login for example phones, laptop or PC. We should either save `sawTutorial` flag per user in some persistent storage (like database) or add `SKIP` button. Saving it on local storage would show this tutorial for same user x number of times. Include cases where storage clean would be needed... and we have even more cases :)
Second thing is preloading images or creating image sized box between slides. Currently modal it's changing it's size after swapping a slide as image is not loading.

Leaving here for track, already left those thoughts on DM.

versx (approved):

clburlison (approved):
very nice improvement. the profile popup is kinda laggy when scrolling on Chrome on macOS. I have **do not** same experience on Safari. Just open the popup switch to another tab, wait 5-10 seconds, switch back, scroll and it stalls before scrolling.

TurtIeSocks:
> very nice improvement. the profile popup is kinda laggy when scrolling on Chrome on macOS. I have the same experience on Safari. Just open the popup switch to another tab, wait 5-10 seconds, switch back, scroll and it stalls before scrolling.

On your M1 I'm assuming? Trying to replicate here on mine but haven't been able to yet.

clburlison:
yes. 91.0.4472.106 arm64 😂  Just tested and can't replicate on my intel x86_64 build. So completely a chromium bug for that platform. Also why I clicked the approve button. Was just noting it in case others saw.

PR #98, opened 2021-06-18
Move location button position with config setting

Petap0w:
Adds a config setting in map section to move the location button to bottom right:
```
  "map": {
[...]
    "locationBtnPosition": "bottomRight",
[...]
  },
```
Capture d’écran 2021-06-18 à 14 18 44

### DIFF
--- a/server/src/configs/default.json
+++ b/server/src/configs/default.json
@@ -34,6 +34,7 @@
     "feedbackLink": "https://forms.gle/wKqWRs9Z7XEAPB7AA",
     "enableStats": false,
     "statsLink": "https://rdmopole2link.net",
+    "locationBtnPosition": "topLeft",
     "startLat": 0,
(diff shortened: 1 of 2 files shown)


TurtIeSocks (changes requested):
Before:
Screen Shot 2021-06-21 at 2 44 37 PM

After:
Screen Shot 2021-06-21 at 2 44 23 PM
 
 Please fix the order.

Petap0w:
> Before:
> Screen Shot 2021-06-21 at 2 44 37 PM
> 
> After:
> Screen Shot 2021-06-21 at 2 44 23 PM
> 
> Please fix the order.

that switch is voluntary: you probably want to zoom in / out more often than submit a feedback or check stats so made them "closer to the thumb distance"

Petap0w:
Mods included in #111

PR #99, opened 2021-06-19
Rename S2cell to ScanCell

Petap0w:
Config setting changes: useFor `"s2cell"` becomes `"scanCell"` in your DB schemas.


TurtIeSocks:
This was a planned change but I'd rather make all of the config changes that actively break people's maps near the end of the beta period, all at once, so people only have to redo their configs a single time. We can leave this open for now but not going to merge it right now.

Petap0w:
Mods included in #111

PR #107, opened 2021-06-23
Provide Docker setup

Fabio1988:


clburlison:
We need to enable the public container package for the GitHub organization if this is something you want to support @TurtIeSocks.

Minor nits:

- This php-Apache base image is 6 months old and 605MB decompressed on disk!! This project doesn't require either php or Apache 
- Using the `node:12-slim` or `node:12-alpine` would be much better, more updated, and much smaller base containers. 
- Should ship this as `docker-compose.example.yml` it's expected that users will make modifications to the file
- Add `docker-compose.yml` to `.gitignore`

Fabio1988:
Thx @clburlison a lot.

Wasn't aware of some requirements. exchanged docker-image.
also renamed `docker-compose.yml` to `docker-compose.example.yml`
removed depends_on
added `docker-compose.yml` to `.gitignore`

I told TurtleSocks about the public container package support. If he wants to support it, would be great :)

docker-compose.example.yml
```diff
@@ -0,0 +1,12 @@
+version: '3.1'
+services:
+  reactmap:
+    image: ghcr.io/WatWowMap/reactmap:main #pr-10
+    container_name: reactmap
+    command: sh -c "sudo yarn start"
```
clburlison, line 6:
```suggestion
    command: "yarn start"
```
Don't need (and can't use `sudo` with alpine)

clburlison (changes requested):
One last change but this looks good. Tested locally. Support for this should be pretty minimal. Can add it into the README later if you want or just leave it as is. Most docker users will be able to see the compose file and do the right thing.

clburlison (approved):

versx (approved):

PR #112, opened 2021-06-24
Don't push forked repos to docker package container

clburlison:
AKA fix the CI/CD failures

External forks can't push to this organization container registry.
Github does this for security reasons. Running the test cases are nice
so they have been moved to the lint workflow.

More information: https://docs.github.com/en/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,28 +1,15 @@
 name: Docker
-on: [push, pull_request]
+on: [push, release]
 
 env:
   REGISTRY: ghcr.io
   IMAGE_NAME: ${{ github.repository }}
 
 jobs:
-  Deploy:
+  Docker:
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        node-version: [12.x]
     steps:
     - uses: actions/checkout@v1
-    - name: Use Node.js ${{ matrix.node-version }}
-      uses: actions/setup-node@v1
-      with:
-        node-version: ${{ matrix.node-version }}
-
-    - name: Test the Testcases
-      run: 
-        npm install -g yarn
-        yarn install
-        yarn build
 
     - name: Log in to the Container registry
       uses: docker/login-action@v1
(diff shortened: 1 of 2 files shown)


clburlison:
Sorry for CI spam. I'm use to doing this with AWS/GCP and not having to deal with external forks. Ready for review/merge. We can't have external forks pushing for security. It provides little benefit so just kill them. All pushes to the `WatWowMap/ReactMap` repo will continue to have tagged docker releases. Plus we have the basic lint tests running in a separate GH action now.

Fabio1988 (approved):

Fabio1988:
So it's defined in your organisation, that you don't allow docker push workflows?
Sorry for the inconvinience, other repos do allow docker builds on push, have not known about that permission restriction. But i think that is only a restriction on your github token? So actually the Docker workflow will only run on your own branches and pushes, pull requests are excluded? I'm fine with that.

Maybe also restrict the github bot for discord, or is it necessary to notify about workflow actions? :)

versx:
Notify all the things

clburlison:
@Fabio1988 Flo is getting around this with RDM by hosting his own build server. Additionally we can't use a personal access token since that environment variable doesn't get used during builds from forks. Tested it for about an hour.

clburlison:
For reference I'm attaching the two relevant org permissions as pictures below.

I even attempted to override the permissions for the token via 
```
permissions:
  package: write
```

and it still sent over a read key. https://github.com/WatWowMap/ReactMap/runs/2909179176?check_suite_focus=true#step:1:13

When using the personal access token that had the correct permissions GH doesn't send over the environment variable. This is to protect organization/user secrets. Imagine some bad actor modifying the GitHub workflow, doing a echo $var_name and that getting sent to the world via a public log.

It's a cool idea but not worth the maintenance of someone running an external server. The benefits simply don't outweigh the time/money/maintenance costs. 

D39E94D6-1E28-462C-816A-2C242C918901

9CE65BBE-4B38-4C2D-87E3-1C5CD87BC777

Fabio1988:
> Imagine some bad actor modifying the GitHub workflow, doing a echo $var_name and that getting sent to the world via a public log.

Totally agree with that, disable it, I think it's good this way :)

TurtIeSocks (approved):

PR #120, opened 2021-06-27
Remove ESLintPlugin

Mygod:
`ESLintPlugin` slows down `yarn start` by 3x, wasting CPU cycles and time while achieving absolutely nothing. We should disable it at least for end users.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -119,10 +119,6 @@ module.exports = (env) => {
           filename: isDevelopment ? '[name].css' : '[name]-[contenthash:8].css',
           chunkFilename: isDevelopment ? '[id].css' : '[id].[contenthash:8].css',
         }),
-        new ESLintPlugin({
-          context: 'src',
-          extensions: ['js', 'jsx'],
-        }),
         new ProgressBarPlugin({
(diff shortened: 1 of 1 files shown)


lenisko:
I wouldn't remove it completely just like you mentioned. Maybe lets add it to GH checks and let devs to execute it using standalone command e.g. `yarn lint` ?

Mygod:
@lenisko That is definitely a better idea. Would you like to implement it?

lenisko:
@Mygod If you won't be able I can take a look. From fast check shouldn't be hard as everything is here: https://github.com/WatWowMap/ReactMap/blob/main/package.json#L9-L21 and what we need is to skip `eslint` task during `webpack` execution by either writing some logic for CLI switch or using two configuration files (for devs and end-users).

TurtIeSocks:
Instead of removing please change the webpack config to:

```js
        ...(isDevelopment ? new ESLintPlugin({
          context: 'src',
          extensions: ['js', 'jsx'],
        }) : []),
```

Mygod:
Sure. You can push directly.

PR #121, opened 2021-06-27
Refine raid timers

Mygod:
* Egg timer tooltip now displays correctly.
* 3:00 will display as `3m` instead of `3m 0s`.
* Display raid duration when both start timer and end timer is on screen.

image
image


TurtIeSocks (changes requested):
Why are we keeping the old function around that is now only providing a simple ternary? 

Please move the ternary into your new function or the Utility class to solve before it hits your function. Then update all of the existing components that are using the old function to only use the new one.

TurtIeSocks (approved):

PR #123, opened 2021-06-29
Add Raid Level Badges

TurtIeSocks:
For Bidoof day!
- Set client side


Mygod:
hi i hate bidoof day

Mygod (approved):
Looks nice. Issues found:

* Switching the option does not apply until refresh (or it is only applied on new gyms to be more precise.)
* Raid level badge circle size does not match gym slots circle size.

PR #133, opened 2021-07-29
Add `All Pokemon Timers` option

lenisko:
Add `All Pokemon Timers` option. Tested. And that's it :)


TurtIeSocks:
Sigh. I was really hoping to avoid this one but I guess there's been enough people asking for it. What do you think of having some sort of warning (see https://github.com/WatWowMap/ReactMap/blob/cde8d83c1a28ac4b2d2362775960d5f4f8c9184b/src/components/layout/drawer/Settings.jsx#L253) when users enable this option to say "Hey, this is going to slow down your map, proceed with caution"?

lenisko:
Tutorial is already mentioning that - but it's up to you :)

`  "tutorialClosing3": "There are lots of aesthetic options available for each category, but keep in mind that the more you add to your map, the higher the chance your map will freeze",`

PR #144, opened 2021-08-11
Fix Hundo on Top

pjorritsma:
This change will make it possible to always some 100 IV (Hundos) on Top of all other pokémon on the map. Before the hundo could hide behind other pokémons.


TurtIeSocks (approved):

lenisko:
I would move to the top only 0 and 100% IVs as those are the only ones in shortcut.
If I'm correct, your change would order them from bottom to the top so 0% IV would become invisible in some cases.

lenisko:
`** lol **` :joy:

PR #155, opened 2021-08-18
Breaking change - Include only 'en' in default `localeSelection`

lenisko:
Include only `english` (as it's a base language) in `default.json` configuration file.
Due to the way configuration is extended those values can't be removed from `Locale Selection` by removing them in `localeSelection` array under `config.json`.

Users who want to include other languages than `english` have to enable them in `config.json` by adding `"localeSelection": ["en", ...]` array.

### DIFF
--- a/server/src/configs/default.json
+++ b/server/src/configs/default.json
@@ -1,7 +1,7 @@
 {
   "interface": "0.0.0.0",
   "port": 8080,
-  "localeSelection": ["en", "de", "pl" ,"es" ,"fr", "it"],
+  "localeSelection": ["en"],
   "devOptions": {
     "enabled": false,
     "graphiql": false,


TurtIeSocks (commented):
I agree with this PR overall, though I think maybe we should add some notes in the readme stating what languages have some sort of support. 

Thoughts on removing language selectors (in Settings and Tutorial) if the array length is 1?

lenisko:
How about we populate `config.example.json` with all supported (translated by someone) languages?

Good idea with selector removal 👌

Another idea would be to use `localeSelection  = config.localeSelection OR default.localeSelection`

TurtIeSocks:
> How about we populate `config.example.json` with all supported (translated by someone) languages?
> 
> Good idea with selector removal 👌
> 
> Another idea would be to use `localeSelection = config.localeSelection OR default.localeSelection` (so when we declare a key in `config.json` we would use it instead merging.

Yeah I think in the long term I'm going to need to change the library we use for merging configs. I'm really not a fan of how it handles arrays.

PR #196, opened 2021-09-24
Reorder search button

Mygod:
Search is not as important as zooming in/out.


TurtIeSocks (changes requested):
Thanks for the PR but can you make it configurable? I'm sure there are others that disagree that Search is less important than zoom.

plinytheelder:
Make it all configurable and lump in moving the location and zoom buttons to the bottom right like all other maps. kthanksbye

PR #204, opened 2021-09-27
Multi-Invasion Support

TurtIeSocks:
- Multiple invasions if running Chuck
- Added knex debug mode
- Couple of forced boolean checks to get rid of weird React bugs


server/src/models/Pokestop.js
```diff
@@ -83,6 +87,16 @@ class Pokestop extends Model {
             .as('incident_expire_timestamp'),
         ])
     }
+    if (dbType === 'chuck') {
+      query.join('incident', 'pokestop.id', 'incident.pokestop_id')
+        .select([
+          '*',
+          'pokestop.id AS id',
+          'incident.id AS incidentId',
+          raw('ROUND(incident.expiration_ms / 1000, 0) AS incident_expire_timestamp'),
```
Mygod, line 96: I personally prefer `FLOOR` instead of `ROUND`. However, it is simply a matter of preference. It is more important to keep things consistent in my opinion, i.e. use either `FLOOR` or `ROUND` everywhere.

PR #221, opened 2021-10-12
Raid Level Filtering

TurtIeSocks:
- Filter by entire raid levels instead of individual Pokemon
- Includes an AND/OR switch client side that decides whether a raid requires both the raid level and the boss to filter or if the raid level filters everything
- New translation key

Examples:
When enabled...
- If Wigglytuff (T3 boss) is selected, it will show no matter what else is selected
- If Raid T3 is selected, it will show all T3s, in addition to Wigglytuff, including bosses that have hatched but are currently unknown

When disabled...
- If Wigglytuff is selected and the Raid T3 is not selected, nothing will show
- If Raid T3 is selected and no raid bosses are selected, it will only show bosses that have hatched but are unknown. 
- If Wigglytuff and Raid T3 is selected, it will show Wigglytuff and unknown raid bosses.

Egg selections will no longer show bosses that have hatched but their species is still unknown.

### DIFF
--- a/server/src/models/Gym.js
+++ b/server/src/models/Gym.js
@@ -82,6 +82,7 @@ class Gym extends Model {
         default: raidBosses.add(gym.split('-')[0]); break
       }
     })
+    if (!onlyOrRaids) raidBosses.add(0)
     const finalTeams = []
     const finalSlots = {
       1: [],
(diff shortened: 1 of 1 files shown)


public/base-locales/de.json
```diff
@@ -385,5 +385,6 @@
   "excludeInvasionMulti": "{{invasion}} ausschließen",
   "messageOfTheDay": "Nachricht des Tages",
   "hasQuestIndicator": "Alternative Farbe für Stops mit Quests",
-  "showArBadge": "Zeige AR Abzeichen"
+  "showArBadge": "Zeige AR Abzeichen",
+  "raidsOr": "Raid Bossen separat filtern"
```
ReuschelCGN, line 389: Raid Bosse separat filtern 🤷‍♂️

PR #238, opened 2021-10-27
Supports Night/Day Weather

TurtIeSocks:
- Removes the last of the local images (except for perms)
- Now uses css to colorize the standard white weather icons
- Checks the hour to determine if it should attempt to display night or day weather icon, if available
- Uses jsx icon now for weather tile


TurtIeSocks:
Could be an interesting idea!

PR #246, opened 2021-11-06
Error Catching

TurtIeSocks:
- Adds some error catching during the startup process to help identify what could possibly go wrong without forcing them to go into the console

### DIFF
--- a/public/base-locales/en.json
+++ b/public/base-locales/en.json
@@ -441,5 +441,6 @@
     "Please try again later."
   ],
   "profile_error": "Must be unique and valid",
-  "team_0": "Harmony"
+  "team_0": "Harmony",
+  "config_error": "There has been a startup error, please report the below issue to your Admin"
 }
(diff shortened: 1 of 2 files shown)


lenisko:
As addition. I can see three groups:
1. Developer errors - those "could" be solved by *cogh*, *cough* tests 😂 and monitored using error monitoring tools (Sentry or Bugsnag have free plans)
2. Server side config - maybe some validation? if it's wrong, many things could go wrong...
3. Client side config/filters - for starter some catch and popup with "Your filters/config are broken. You can download them <here> and restore defaults using button below."

TurtIeSocks:
Server side could definitely use some TDD, client side tests are pretty overwhelming at this point though... TypeScript would also make a big difference. 

Validation for the config would definitely help..._TypeScript_

I was thinking for client side errors I could add a GraphQL endpoint that would be for error logging. Then add try/catch blocks in the more complex/logic heavy components to report any errors to that endpoint, that way they're then logged and available for admins to see. 

I'm guessing that most users aren't reporting client side errors so it could help quite a bit, if I actually committed to adding try/catch in enough components. We get the user perms in the graphql endpoint as well so errors could be replicated more easily.

TurtIeSocks:
Going to come back to this when I'm ready to do a full error catching implementation.

PR #255, opened 2021-11-12
Translate webhook management in french

na-ji:
Hello,

I quickly translated every keys I found used in the webhook management. In case another translator is wondering: there are still a few things not translated, but they are coming from Poracle's side (like tracked things descriptions).

Thanks.


TurtIeSocks (approved):

TurtIeSocks:
Thank you for the locale contribution! It's greatly appreciated.

PR #295, opened 2021-12-31
All the locales

TurtIeSocks:
- Fills in the remaining of the locales that pogo assets are available for
- Adds Swedish locale


lenisko:
I'll make changes after new year. Those machine translations are in 80% wrong :joy: but at least we know which keys need to be updated 🦾

PR #300, opened 2022-01-02
Add Quest Set Toggle

Petap0w:
Adding an option to filter quest set on map between "with AR", "without AR" or both.

Known bug: when switching directly from with to without or vice versa the quests "on screen" don't change, you have to pan away and come back for refresh... TurtIe might know how to fix this :)

https://user-images.githubusercontent.com/47784174/147892230-d3271ae3-6ea2-49f8-a550-94c7ff94cae8.mov


TurtIeSocks:
I will review this later in the week when I have more time to sit down and figure out the bad state. Assuming your second commit makes this optional for those of us not scanning dual quests?

TurtIeSocks (approved):

PR #301, opened 2022-01-03
Modify TileServers selection

Petap0w:
- Moves tileServers from default.json to config.example.json
- Adds config settings to make Auto tile server selection optional / forced.


TurtIeSocks:
I acknowledge that this is a problem, and not just for the tileserver section. I am, however, going to close this as I have an in progress PR to completely revamp the config to be more approachable and to not have this undesirable behavior we're seeing with both objects and arrays. Thanks!

PR #302, opened 2022-01-03
Make tutorial optional

Petap0w:
- Make tutorial prompt forced / not forced
- Allow / disable the tutorial option


TurtIeSocks (changes requested):
This PR is unnecessarily complicated. The disabling part is fine (though I'm crying internally) but I think if an admin sets `forceTutorial: false` then this should simply be setting the tutorial prop in the strategies (when writing a new user to the db) to be true by default. This simplifies the frontend code and reduces the unnecessary logic. 

See this line:
https://github.com/WatWowMap/ReactMap/blob/def919cdedbd3b903666e01750ec2f5717b49d55/server/src/strategies/discord.js#L71

Change to:
```js
.insertAndFetch({ discordId: user.id, strategy: 'discord', tutorial: !config.map.forceTutorial })
```

And voila, you have accomplished the same thing, while keeping the user table in sync. Just make sure to do it for every other "official" strategy too ;)

Petap0w:
> This PR is unnecessarily complicated. The disabling part is fine (though I'm crying internally) but I think if an admin sets `forceTutorial: false` then this should simply be setting the tutorial prop in the strategies (when writing a new user to the db) to be true by default. This simplifies the frontend code and reduces the unnecessary logic.
> 
> See this line:
> 
> https://github.com/WatWowMap/ReactMap/blob/def919cdedbd3b903666e01750ec2f5717b49d55/server/src/strategies/discord.js#L71
> 
> Change to:
> 
> ```js
> .insertAndFetch({ discordId: user.id, strategy: 'discord', tutorial: !config.map.forceTutorial })
> ```
> 
> And voila, you have accomplished the same thing, while keeping the user table in sync. Just make sure to do it for every other "official" strategy too ;)

Yes right, did that part to sync the user db, but where do you want me to set the initial state of `tutorial` then?

TurtIeSocks:
> > This PR is unnecessarily complicated. The disabling part is fine (though I'm crying internally) but I think if an admin sets `forceTutorial: false` then this should simply be setting the tutorial prop in the strategies (when writing a new user to the db) to be true by default. This simplifies the frontend code and reduces the unnecessary logic.
> > See this line:
> > https://github.com/WatWowMap/ReactMap/blob/def919cdedbd3b903666e01750ec2f5717b49d55/server/src/strategies/discord.js#L71
> > 
> > Change to:
> > ```js
> > .insertAndFetch({ discordId: user.id, strategy: 'discord', tutorial: !config.map.forceTutorial })
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > And voila, you have accomplished the same thing, while keeping the user table in sync. Just make sure to do it for every other "official" strategy too ;)
> 
> Yes right, did that part to sync the user db, but where do you want me to set the initial state of `tutorial` then?

I'm not following? This seems to work perfectly now, both the forcing and disabling settings, with the changes you made.

Petap0w:
Nevermind, I fixed the issue with last commit. Should be good to go now.

TurtIeSocks (approved):

PR #304, opened 2022-01-03
Make profile (perms) panel optional

Petap0w:


src/components/layout/Nav.jsx
```diff
@@ -123,7 +123,7 @@ export default function Nav({
           setDonorPage={setDonorPage}
         />
       )}
-      {userProfile ? (
+      {(enableUserProfile && userProfile) ? (
```
TurtIeSocks, line 126: Is this check necessary @Petap0w ? I don't think it should be if the profile button has been removed entirely.

TurtIeSocks (changes requested):
You also need to remove the profile button from the tutorial component when disabled.

PR #307, opened 2022-01-07
Fix some locale errors

Petap0w:


Petap0w:
Will PR in all-the-locales instead

PR #325, opened 2022-01-15
Generate Area.json on startup

pjorritsma:
**Work in progress**

Have you ever wished you could generate area.json more easily?

This is based on code from https://gist.github.com/moriakaice

How does it work Poracle:

1. Copy Geofence file from Poracle to Public/Geofence/geofence.json
2. Run Yarn create-area-poracle 
3. Area.json will be generated in src/configs/area.json 

ToDo:
- Make config option to support GeofenceToGeoJson or PoracleToGeoJson,
- Make generation of area.json automatic on startup,
- Update code (some line can be reduce


docker-compose.example.yml
```diff
@@ -7,7 +8,10 @@ services:
     restart: unless-stopped
     volumes:
       - ./server/src/configs/areas.json:/home/node/server/src/configs/areas.json
-      - ./server/src/configs/local.json:/home/node/server/src/local/config.json
+      - ./server/src/configs/local.json:/home/node/server/src/local/local.json
+      - ./geofence/:/home/node/public/geofence
```
TurtIeSocks, line 12: Why are we copying the geofence into the public folder? It's not used there...

public/geofence/geofence.json
```diff
@@ -0,0 +1,181 @@
+[
```
TurtIeSocks, line 1: What purpose does this file add?

server/scripts/geofenceToGeoJSON.js
```diff
@@ -0,0 +1,70 @@
+/**
```
TurtIeSocks, line 1: Is this format commonly used in the pogo community? I thought we mostly just Poracle and GeoJson.

server/scripts/poracleToGeoJSON.js
```diff
@@ -0,0 +1,68 @@
...
+    console.error(err);
+    return;
+  }
+  const inGeoJSON = JSON.parse(data);
+  if (inGeoJSON.length === 0) {
+    // eslint-disable-next-line no-console
+    console.error('Failed to parse poracle geofence file');
+    return;
+  }
+  // eslint-disable-next-line no-plusplus
```
TurtIeSocks, line 37: For ESlint console warnings, you can disable them for the whole file instead of line by line, but please for the rest of them, fix the code instead of skipping the rule. Please also remove `;` as they aren't used in this codebase.

TurtIeSocks (changes requested):
I don't think we should place the geofences into the public folder as they don't really make sense there. i18n has to read from the translations in that folder, is why it exists for locales. 

I would suggest adding a `server/src/configs/geofence/*` to the .gitignore file, since everything related to areas goes in the configs folder anyway. 

I'm not entirely convinced that the geofence=>areas.json is necessary for this project? Do any other active pogo projects use that format? 

The poracle=>areas.json script is fine, just need to fix the eslint issues and the input folder.

Thanks for the pr!

public/geofence/geofence.json
pjorritsma: Example file for people like area.json

docker-compose.example.yml
pjorritsma: Will push update for it.

server/scripts/geofenceToGeoJSON.js
pjorritsma: No, mostly format will be Poracle and will remove this one. So we get one method.

server/scripts/poracleToGeoJSON.js
pjorritsma: Read the information about it and add ESLint to code and remove the ; from the code.

PR #327, opened 2022-01-16
Additional Version Error Checking

TurtIeSocks:
This has turned into a general bug fix PR:
- Better update checking for non-docker users
- Fix update checking in general for docker users
- Add better error catching for GraphQL Validation Errors for clarity
- Ohbem init log
- Support for custom strategy types
- Session init checks for webhooks
- Poracle profile editing check
- Gym/Pokestop recentering
- Fix DevicePoly for MariaDb users
- DevicePoly route color memo check
- Fix some bool checks in the Pokemon popup


Fabio1988:
So in this CheckForUpdate.js you only compare content of package.json version.
If someone running a PR docker image the fetching of package.json can not be successful, because github.ref points to PR#327. But for develop and main this does work.

But maybe we should rework it without using package.json

Using following commands:

`git ls-remote $REPO_URL refs/heads/$branch_name`

example:

> git ls-remote https://github.com/WatWowMap/ReactMap/ refs/heads/main
> cab92c5ec6a05ecf956b42daff7633d1c034e48f	refs/heads/main
> 
> git ls-remote https://github.com/WatWowMap/ReactMap/ refs/heads/develop
> 8d2272ca8ec335b7e0221e654622882303b75db8	refs/heads/develop

If someone is running without docker it's possible to fetch git sha with the following command:
`git rev-parse HEAD`

Fabio1988:
In docker $branch_name would be written in .gitref file

running withoud docker it would need the `git branch --show-current` command to fetch branch_name

server/src/services/checkForUpdates.js
```diff
@@ -1,32 +1,40 @@
...
-exec('git branch --show-current', async (err, stdout) => {
-  try {
-    if (err || typeof stdout !== 'string') {
-      throw new Error('Unable to get current branch', err.message)
+    if (!gitRef && (err || typeof stdout !== 'string' || !stdout.trim())) {
+      throw new Error('Unable to get current branch', err)
     }
-    const { version } = await Fetch.json(`https://raw.githubusercontent.com/WatWowMap/ReactMap/${stdout.trim()}/package.json`)
+    const branch = typeof gitRef === 'string' && gitRef.trim()
+      ? gitRef.trim()
```
Fabio1988, line 13: for branch name you need to remove prefix. content of the `.gitref`-file: `refs/heads/addl-version-checking`

Fabio1988 (changes requested):

PR #335, opened 2022-01-21
Add 'Reset filters' dialog

kamieniarz:
Tested locally - everything works as intended. I hope the code is clean enough ;)

image

### DIFF
--- a/src/components/layout/Nav.jsx
+++ b/src/components/layout/Nav.jsx
@@ -39,8 +39,7 @@ export default function Nav({
   const setUserProfile = useStatic(state => state.setUserProfile)
   const feedback = useStatic(state => state.feedback)
   const setFeedback = useStatic(state => state.setFeedback)
-  const resetfilters = useStatic(state => state.resetfilters)
-  const setResetFilters = useStatic(state => state.setResetFilters)
+  const resetFilters = useStatic(state => state.resetFilters)
 
   const filters = useStore(state => state.filters)
   const setFilters = useStore(state => state.setFilters)
@@ -211,10 +210,10 @@ export default function Nav({
         <Feedback link={config.feedbackLink} setFeedback={setFeedback} />
       </Dialog>
       <Dialog
-        open={resetfilters}
+        open={resetFilters}
         maxWidth={isMobile ? 'sm' : 'xs'}
       >
-        <ResetFilters setResetFilters={setResetFilters} />
+        <ResetFilters />
       </Dialog>
       <Snackbar
         open={Boolean(webhookAlert.open)}
--- a/src/components/layout/Nav.jsx
+++ b/src/components/layout/Nav.jsx
@@ -18,6 +18,7 @@ import Search from './dialogs/Search'
 import Motd from './dialogs/Motd'
 import DonorPage from './dialogs/DonorPage'
 import Feedback from './dialogs/Feedback'
+import ResetFilters from './dialogs/ResetFilters'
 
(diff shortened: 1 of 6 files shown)


src/components/layout/Nav.jsx
```diff
@@ -38,6 +39,8 @@ export default function Nav({
   const setUserProfile = useStatic(state => state.setUserProfile)
   const feedback = useStatic(state => state.feedback)
   const setFeedback = useStatic(state => state.setFeedback)
+  const resetfilters = useStatic(state => state.resetfilters)
+  const setResetFilters = useStatic(state => state.setResetFilters)
```
TurtIeSocks, line 43: Leave `resetFilters` here, but I would move `const setResetFilters = useStatic(state => state.setResetFilters)` into the `ResetFilters.jsx` file itself. Since you're not using it in `Nav.jsx`, other than passing it through to the dialog, may as well move it to the dialog itself, particularly since then you will have a "clean" component, without any props passed through.

src/components/layout/Nav.jsx
```diff
@@ -207,6 +210,12 @@ export default function Nav({
       >
         <Feedback link={config.feedbackLink} setFeedback={setFeedback} />
       </Dialog>
+      <Dialog
+        open={resetfilters}
+        maxWidth={isMobile ? 'sm' : 'xs'}
+      >
+        <ResetFilters link={config.resetfiltersLink} setResetFilters={setResetFilters} />
```
TurtIeSocks, line 217: It doesn't look like you're using `link={config.resetfiltersLink} ` anywhere in the `ResetFilters.jsx` dialog? Did you forget to add something? (If not, you can just remove it entirely)

src/components/layout/dialogs/ResetFilters.jsx
```diff
@@ -0,0 +1,42 @@
...
+        <br />
+        <Divider />
+        <br />
+        <Typography align="center">
+          <Button
+            style={{ minWidth: 100 }}
+            variant="contained"
+            color="primary"
+            size="small"
+            onClick={() => { setRedirect(true); setResetFilters(false) }}
```
TurtIeSocks, line 32: Add some lines here instead of the `;`
```
{
setRedirect(true)
setResetFilters(false)
}
```
etc

TurtIeSocks (changes requested):
Great first PR, I just have a few minor change requests. 

Thank you for the contribution!

src/components/layout/Nav.jsx
lenisko: @kamieniarz 

* move `const setResetFilters = useStatic(state => state.setResetFilters)` line to `ResetFilters.jsx`
* replace `export default function ResetFilters({ setResetFilters })` with `export default function ResetFilters()` in `ResetFilters.jsx`
* replace `<ResetFilters setResetFilters={setResetFilters} />` with `<ResetFilters />` in `Nav.js`

Haven't checked myself but should be everything you need.

src/components/layout/dialogs/ResetFilters.jsx
```diff
@@ -0,0 +1,48 @@
...
+        <Divider />
+        <br />
+        <Typography align="center">
+          <Button
+            style={{ minWidth: 100 }}
+            variant="contained"
+            color="primary"
+            size="small"
+            onClick={() => 
+				{
```
lenisko, line 34: Project is using spaces instead tabs. Replace them with spaces or run linter.

TurtIeSocks (approved):

kamieniarz:
Thanks for the fixes and merging ;) I was away today so I couldn't commit them

PR #349, opened 2022-01-24
Raids Quick Selection

TurtIeSocks:
- Add drop down to main sidebar to quickly select different raid tiers
- Removes raidsOr, this was confusing to users and was only really meant to be used with the webhook manager stuff
- Untested at the moment


kamieniarz:
I've been testing this PR for 2 days and sometimes when using these filters map shows Gyms without any raid, these Gyms seem to be random. There's no such issue with `All`

PR #350, opened 2022-01-24
new loading screen translations

ReuschelCGN:


TurtIeSocks (changes requested):
@ReuschelCGN can you also add translations to the two additional fields added in this commit:
https://github.com/WatWowMap/ReactMap/pull/344/commits/2239e750dcb4debbc2947a8004b89f5ae84ca701

Due to these fields being loaded _before_ the translations JSONs, you must add them in these files compared to the usual JSON.

Issue #360, opened 2022-01-27
`Edit Gym Badge` menu doesn't work right when `Gym Badges` is disabled


kamieniarz:
`Edit Gym Badge` doesn't contain actual badge data when `Gym Badges` is disabled. I think this option should be removed from Gym popup when badges are disabled or the other way - keep the data always available.

PR #371, opened 2022-01-28
Grey out gyms not updated since x days

Petap0w:
- Will turn to team Harmony gyms not updated since more than `gymValidDataLimit` days from `config` > `api` (default `30` days)
- Removes gym `Defender:` and `Total CP:` from popup footer when over this limit
- Removes `Exclude Team` from those gym popup `...` menu

Sides:
- Removes `Defender:` placeholder when gym has no team
- Removes `Edit Gym Badge` from gym popup `...` menu when that filter isn't enabled in `Menu` > `Gyms` (Issue #360)
- Moves down `10px` all popups `...` menu so people are less likely to hide an item when trying to hit the X

greyoldgym


plinytheelder:
can you make the increment smaller? like hours or minutes? Some of us quest scan more than we actively spawn scan, and don't want to give an incorrect indication for team/slots available.

Petap0w:
you can use `gymValidDataLimit: 1` for 1 day, or `gymValidDataLimit: 0.041` for 1 hour, etc.

kamieniarz:
Weird thing - I disabled Harmony Gyms in advanced settings but I can still see them when both Raid and All Gyms switches are enabled 🤔 No idea if that's this PR thing but I noticed it during while testing it

TurtIeSocks:
@Petap0w I made some changes, let me know if everything is still working as anticipated. If so, this is ready to merge. Thanks for the PR!

PR #379, opened 2022-01-28
Generate area.json on startup

pjorritsma:
Have you ever wished you could generate area.json more **easily?**

With this Pull Request a area.json file will be generate in your config folder. Why do you need area.json? With area.json in you config folder it's possible to show Polygons on you Map (Let users see the scan area or just for review). You can find Polygons under Menu > Scan Area > Show Polygons. 

Requirements:
- Geofence Folder in Config folder
- Geofence.json file with the following format:

Example geofence.json:
`[
    {
        "name": "New York",
        "color": "#6CB1E1",
        "id": 0,
        "path": [
            [
                40.6758460704592,
                -74.0183327629713
            ]
        ]
    }
]
`
**This is the same format as Poracle and other tools.**

_The JS script used in this PR is orginal from https://gist.github.com/moriakaice_

**How can I run this locally:**

1. Copy Geofence file from Poracle to Config/Geofence/geofence.json
2. Run Yarn create-area

A Area.json will be generated in src/configs/area.json


pjorritsma (commented):
Thanks for the cleanup and fix :)  runs smooth!

PR #382, opened 2022-01-30
Enable new Pokemon by default

Mygod:
ok thanks


TurtIeSocks (changes requested):
Please test these changes as I don't think this is going to do what you think it will, thanks!

TurtIeSocks:
Closing due to inactivity.

PR #383, opened 2022-02-03
Introducing scanNext & scanZone

Petap0w:
This PR adds a new feature for an "on demand scan" directly managed from the UI on the map.
When enabled, and connected to the scanner API (for RDM see PR297, MAD supported as well), new floating buttons allow qualified users (perms / role based) to request a scan on a chosen location from the map.

`scanNext` - There are 3 modes available:
- `S` for a single location scan,
- `M` for a group of 7 nearby locations (where's that wild Pokémon I saw on the radar and what's its IVs?) - RDM only,
- and `XL` for a large coverage of 7 gym-range locations (looking for a raid, TR invasions, or adding gym badges to my collection) - RDM only.

`scanZone` - You can select the size directly with a slider from the popup menu.

Additionally, this can be restricted to certain `areas` from your `areas.json` if you don't want your users to request scans all over the world.

New config options:
```
  "scanner": {
    "backendConfig": {
      "platform": "rdm/mad",                -> scanner backend type
      "apiEndpoint": "http://ip:port/api",  -> scanner API endpoint
      "apiUsername": "username",            -> authorized API user credentials
      "apiPassword": "password",
      "queueRefreshInterval": 5             -> Minimum interval to request queue from scanner
    },
    "scanNext": {
      "enabled": false,                     -> to enable the scanNext floating button and feature
      "showScanCount": false,               -> Show number of scan request in popup
      "showScanQueue": false,               -> Show current queue in popup
      "scanNextInstance": "scanNext",       -> the scanner instance to which the requests will be made (RDM)
      "scanNextDevice": "Device01",         -> the scanner device to which the requests will be made (MAD)
      "scanNextSleeptime": 5,               -> MAD sleeptime option
      "scanNextAreaRestriction": [],        -> allowed areas for the scans (needs a valid areas.json with name matching)
      "discordRoles": [],                   -> Discord roles that will be allowed to use scanNext
      "telegramGroups": []                  -> Telegram group that will be allowed to use scanNext
    },
    "scanZone": {
      "enabled": false,                     -> to enable the scanZone floating button and feature (RDM only)
      "showScanCount": false,               -> Show number of scan request in popup
      "showScanQueue": false,               -> Show current queue in popup
      "scanZoneMaxSize": 10,                -> the maximum size allowed for the scan zone area
      "advancedScanZoneOptions": false,     -> enables advanced options in popup (spacing and radius)
      "scanZoneRadius": { "pokemon": 70, "gym": 750 },     -> default circle sizes
      "scanZoneSpacing": 1,                 -> default spacing coefficient
      "scanZoneInstance": "scanZone",       -> the scanner instance to which the requests will be made
      "scanZoneAreaRestriction": [],        -> allowed areas for the scans (needs a valid areas.json with name matching)
      "discordRoles": [],                   -> Discord roles that will be allowed to use scanZone
      "telegramGroups": []                  -> Telegram group that will be allowed to use scanZone
    }
  },
```

_Requires `yarn install` after pulling to install new dependencies_

https://user-images.githubusercontent.com/47784174/152394031-86e62ad9-2f36-4107-9f09-fab6043a945d.mp4

https://user-images.githubusercontent.com/47784174/152694261-316f3a16-8c12-451a-8904-f05fc6449f71.mp4

https://user-images.githubusercontent.com/47784174/152394078-f01d9b84-3899-440e-894b-a328dff86576.mp4


lenisko:
> Noteworthy: It needs additional securing with an API secret, for which I need TurtIe's input.

What do you mean? I guess you need only RDM username/password for API access and everything should be handled by server side of RM not Front.

I was brainstorming about #PR297 a bit with @Fabio1988, overall idea looks okay.

Few things to note. We saw two ways to handle this:
1. Borrow device from working instance. You are slowing down some instance/instances. How about limits? How about splitting those "borrows" per instances? What about potential jump distances cooldowns?
2. Provide instances just for those cases.

For both cases, you need a visible queue. Users should know that they will wait approx ~ minutes before getting a result.

As enchantment, I guess users could have the ability to loop checks on the provided area for some time. This way those results make more sense, unlike a one-time check - but that makes a problem of cooldowns, and the best way would be to jump between those areas.

Overall I like an idea, I mean it's a bit useless at this stage, but I see a potential. Still, I see a need for a huge amount of logic, as for now it's just "a nice thing".

TurtIeSocks:
Cool! I will have more feedback once I can sit down and test and go through the code a bit more. Great work!

server/src/routes/rootRouter.js
```diff
@@ -74,7 +74,7 @@ rootRouter.post('/scanNext', async (req, res) => {
           { lat: parseFloat(coord[0].toFixed(5)), lon: parseFloat(coord[1].toFixed(5)) }))),
       }
       console.log(`[scanNext] Request to scan new location by ${scanData.username}${scanData.userId ? ` (${scanData.userId})` : ''} - type ${scanData.scanNextType}: ${scanData.scanNextLocation[0].toFixed(5)},${scanData.scanNextLocation[0].toFixed(5)}`)
-      Axios.get(`${config.scanner.apiEndpoint}set_data`, {
+      Axios.get(`${config.scanner.apiEndpoint}/set_data`, {
```
Fabio1988, line 77: You now have the slash `/` in both? In config and get itself

kamieniarz:
Ok so I've been using this PR for a few days and since it's a quite big feature and it changes a lot for map users I have some ideas that are worth looking into. But at first I'd like to say I found no bugs that should be handled in here - the only thing that bothers me is handling empty areas in RDM. It simply takes ages while MAD simply uses sleeptime and it takes same time as any other areas. But if it's a RDM thing then I guess it should be fixed there ;)

1. Scan Queue - some kind of a queue that could appear before accepting scan location/area + estimated waiting time for that queue. After enabling scan queue could also be visible with ~min time estimation.
2. Quest scan - nothing to explain here I think
3. Scan location/area cooldown per access role or maybe just a global one but I think admin shouldn't be affected by that
4. Current scan status - some kind of a status for a current scan, especially when there're multiple locations for M, XL locations and areas. Right now admin can see devices jumping around but users proly won't have devices enabled so it just would be good to show them the status and message that scan was completed
5. Log channel - sth like log channel for login attempts we already have but simply with a list: user + coords and whatever should be usable. Just a nice and handy log with scanNext and scanZone logs
6. Let using mad and rdm simultaneously, eg. mad for scanNext, rdm for scanZone
7. A tutorial page with an explanation for all these new functions

Anyway great job @Petap0w, I'm looking forward to test more and more ;)

Pogodenhelder:
Running this for 2 days now, and it looks awesome, i can say this works great.

TurtIeSocks:
Oh boy. This PR is now 1100 lines 😭

TurtIeSocks:
Updated with latest develop.

nileplumb:
Works well, but would consider adding extra documentation explicitly stating the user credentials for rdm are login ones and required priviledges.

Also a note on potential conflicts with firewall/nginx/apache/cloudflare blocking the request the administrator may need to address

fosJoddie:
tested against MAD, it does the trick! But it also spams my logs complaining that it can't get the queue (as MAD doesn't have one (yet) )
```
[scannerApi] Getting queue for method scanNext
[scannerApi] There was a problem processing that scanner request
```

Issue #366, opened 2022-01-28
Nests - Add minimum spawns/h filter


lenisko:
Yap

TurtIeSocks:
Solved in #386

PR #386, opened 2022-02-04
Average Nest Slider

TurtIeSocks:
- Adds a slider in the nest section to filter avg number of spawns
- Renames withSliders to Pokemon, because it only serves Pokemon really
- New config key, doesn't necessarily set defaults, sets the slider min/max as well
- translation key
- Addresses #366


lenisko:
Perfect !

PR #391, opened 2022-02-08
Telegram auth fix

Ninjasoturi:
Telegram auth was letting anyone log in as long as a valid chat was configured in a permission's role. This should make it behave as intended. 

I'm not a JS programmer, so please make sure this edit makes sense.

I also changed `image-rendering` from `crisp-edges` to `auto` to make all the icons look nicer


TurtIeSocks:
Thank you for the contribution! 

Can you clarify a bit what the expected vs actual behavior is? As far as I can tell, and my brief testing, this produces the same results? I don't use TG in my daily setup so I'm not 100% familiar with the ins and outs. 

Thanks!

Ninjasoturi:
Sure!


Config:
```
    "strategies": [
      {
        "name": "telegram",
        "type": "telegram",
        "enabled": true,
        "botToken": "000:aaa",
        "groups": ["group A","group B","group C"]
      }
    ],
    "areaRestrictions": [],
    "excludeFromTutorial": [],
    "alwaysEnabledPerms": [],
    "perms": {
      "map": {
        "enabled": true,
        "roles": ["group A", "group B","group C"]
      },
      "pokemon": {
        "enabled": true,
        "roles": ["group A"]
      },
      "iv": {
        "enabled": true,
        "roles": ["group B"]
      },
      "pvp": {
        "enabled": true,
        "roles": ["group C"]
      },
```

Expected behavior is that user attempting to log in has to be a part of one or more of the groups specified in `groups`, and after log in gets limited to the things specified for those groups.

Current behavior is that any Telegram user can access the map and any enabled feature with a valid group ID, even if they were in only one of the groups.

So basically the previous `groupInfo` const was filled with every group id no matter the response from Telegram API.

jfberry:
This is because:
```
const groupInfo = await Promise.all(strategyConfig.groups.filter(async group => {
```
in the original would always get everything (the promise returned from the async group would be truthy)

PR #396, opened 2022-02-14
Add unicode letters

kamieniarz:


lenisko:
You can optimize this regex.

`\p{L}` - matches any kind of letter from any language.
In this case you don't need `A-Za-zÀ-ÖØ-öø-ÿ` anymore.

TurtIeSocks (approved):

PR #405, opened 2022-02-24
Read gymBattles flag from Poracle and act accordingly

Ninjasoturi:
Requires Poracle PR676, might work without it also, I didn't test it.

Hides In battle switch from gym alarm selection and based on the value received alters the quick add gym alarm -functionality.
I guess it's still possible for users to inject the `battle_changes` flag from frontend, but that isn't a major concern.

And again a warning, I still learning JS, so make sure this makes sense!

### DIFF
--- a/server/src/services/api/resolveQuickHook.js
+++ b/server/src/services/api/resolveQuickHook.js
@@ -2,9 +2,9 @@
 const fetchJson = require('./fetchJson')
 const config = require('../config')
 
-const getWildCards = (category) => {
+const getWildCards = (category, gymBattles) => {
   switch (category) {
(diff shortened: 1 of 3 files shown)


TurtIeSocks:
Great PR! At first glance, my only ask, at first glance, is that we wait to see if Jabes bumps the patch version as requested on PR 676, then we do so here as well during the version check step. I will test later today, thank you!

TurtIeSocks:
@Ninjasoturi I added the version check and removed the unused prop added to the client settings, going to merge now.

TurtIeSocks (approved):

PR #409, opened 2022-03-02
Various Missing Locales

TurtIeSocks:
Tagging @Fabio1988 @lenisko @Petap0w @pjorritsma to check the machine translations :)


ReuschelCGN:
think the german translations are fine

PR #432, opened 2022-03-27
Poracle Tracked - fix width & remove offset to fit more rows

lenisko:


TurtIeSocks:
Screen Shot 2022-03-30 at 6 48 18 PM

I understand the height adjustment, but why the width adjustment on the one side?

lenisko:
Here's develop without any changes

2022-03-31_03-40-36-firefox

TurtIeSocks:
Going to close this since the changes are also present in #433 and are being worked on there.

PR #473, opened 2022-04-29
Update default.json

Dewpg:
Updated to identify unneeded section for RDM users.


TurtIeSocks:
Thanks for the contribution but this is already documented in the Wiki and I would like to keep all of the documentation there instead of mixing it in the `default.json`.

PR #493, opened 2022-05-17
Bring Back Raid Filter

TurtIeSocks:
Brings back the OR based raid tier filter for the non-webhook menu.


TurtIeSocks:
Revisit later

PR #503, opened 2022-05-27
Improved areas.example.json

Ninjasoturi:
Improved the format in the example file to better display the order of longitude and latitude, and that the first and last coordinates of the fence need to be the same.

### DIFF
--- a/server/src/configs/areas.example.json
+++ b/server/src/configs/areas.example.json
@@ -7,8 +7,20 @@
         "coordinates": [
           [
             [
-              -100,
-              4
+              lon1,
+              lat1
+            ],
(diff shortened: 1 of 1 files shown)


pjorritsma:
Why should this be improved, I don't have any problem areas on the map without first and last. If we change (Merge) this also the create-area should be improved.

Ninjasoturi:
The polygons are drawn on the map correctly even if the first and the last coordinates aren't the same. But when using `areaRestrictions`, having the wrong coordinate format will result in mysql error `Invalid GIS data provided to function st_geomfromtext` (at least with MySQL 8).

PR #508, opened 2022-06-05
Change `disabled` to `all` in Raid Quick Selector

kamieniarz:
Since default value actually means `all` it makes no sense to have it marked as `disabled`


TurtIeSocks:
I agree that `disabled` is perhaps not the best term but I don't think `all` is better. The reason I went with `disabled` is that it was meant to indicate that the quick filter itself is specifically disabled and is now relying on the advanced filter, which doesn't really align with `all`. I will put it to a vote in the Discord what term would be best.

TurtIeSocks:
Closing this following the results of the poll

PR #512, opened 2022-06-19
Available Slots Compatibility

TurtIeSocks:
- Updates SQL query to use a dynamic column name based on a startup check
- Maintains backwards compatibility for older RDM clients that are still using availble_slots


Fabio1988:
This misspelling is fixed in rdm master since december.... I kept old column for compatibility. 

Why not simple use it and remove the magic to keep it clean? Someone who runs RDM master from december is save. Someone who's not running this state can also run RM from an older state 🤣

TurtIeSocks:
More backwards compatibility = less dumb questions to answer later 😄

PR #547, opened 2022-08-03
Support MAD Layer Column

TurtIeSocks:
- Checks to see whether the new `layer` column exists
- Selects it as `with_ar` for easy compatibility with existing code
- Backwards compatible

**Need to test with MAD and RDM**


crhbetz:
Can confirm this working as intended on the latest MAD async branch. Unfortunately I don't run any main instances anymore, so not able to confirm backwards compatability.

PR #551, opened 2022-08-09
Added a grass icon for Pokemon found in nearby cells

Ninjasoturi:
A marker for for mons found in nearby cells so users can see right away that their location isn't accurate.
I didn't quite understand the current calculation for icon placement when accounting for the pokemon icon offsetX so I just set the position to fixed 10%.

### DIFF
--- a/src/components/markers/pokemon.jsx
+++ b/src/components/markers/pokemon.jsx
@@ -50,6 +50,18 @@ export const fancyMarker = (
           width: size,
         }}
       />
+      {pkmn.seen_type === 'nearby_cell' && (
+        <img
+          src={Icons.getMisc('grass')}
+          alt='nearby_cell'
+          style={{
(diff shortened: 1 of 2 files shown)


TurtIeSocks (commented):
Thanks for the PR!

Logic seems fine, I want to think on this a little bit though to think if there's an alternative before committing to making cell mons use the "fancyMarker", since it's a little bit slower and Pokemon rendering is already on the slower side.

Ninjasoturi:
Sure! I thought about that too, but figured that usually you don't have that many cell mons passing the filters that it'd make that big a difference. Or at least I don't 😃

TurtIeSocks (approved):

TurtIeSocks:
Eh after testing, performance seems fine. Thanks for the cool PR!

PR #566, opened 2022-09-12
Fix MAD devices not showing up on map correctly

pegasy:
The X and Y coords are swapped currently for MAD devices. This is because X is mapped to latitude and Y is mapped to longitude which is incorrect in this case.  Some more info: https://gdal.org/tutorials/osr_api_tut.html#crs-and-axis-order This is a very confusing topic overall when working with geometries and both versions exsists (latitude, longitude / longitude, latitude) depending on the libraries used and also db engines.

### DIFF
--- a/server/src/models/Device.js
+++ b/server/src/models/Device.js
@@ -19,8 +19,8 @@ module.exports = class Device extends Model {
           'settings_area.name AS instance_name',
           'mode AS type',
           raw('UNIX_TIMESTAMP(lastProtoDateTime)').as('last_seen'),
-          raw('X(currentPos)').as('last_lat'),
-          raw('Y(currentPos)').as('last_lon'),
+          raw('Y(currentPos)').as('last_lat'),
+          raw('X(currentPos)').as('last_lon'),
           raw(true).as('isMad'),
         ])
     } else {


TurtIeSocks:
🤔 it seems unlikely to me that this wasn't previously noticed by anyone else running MAD? When I had a MAD DB available for testing, the current code produced the correct positions.

pegasy:
I do agree that it seems odd I would be the only one with this issue. 
I'm currently running MAD async branch, the devices did work when I was on master branch. 
[This](https://cdn.discordapp.com/attachments/217602699783438336/1018970786893156514/unknown.png) picture show how the trs_status (from master branch) view behaved and [this](https://cdn.discordapp.com/attachments/217602699783438336/1018969325501169877/chrome_UBHfs7Jcgd.png) shows how trs_status (from async branch).
I really don't understand why trs_status seems to be the only table where the naitive Point db type is used, all other tables uses latitude double, longitude double.

TurtIeSocks:
Thanks for the screenshots, that's interesting. Is there a chance this is unintentional for MAD async branch? Maybe @fosJoddie can provide some commentary. I believe @crhbetz is also running the async branch, can you confirm this is an issue for you as well?

pegasy:
Current workaround I'm using is to create an empty db and add some views and moved the device part of the db config to point at the new db, seems to be working just fine :)

sql to create views (pokemon view seems to be needed because reactmap uses that to determine db type (mad/rdm)?, what is "type":"mad" even used for?)
```
create view pokemon as
select * from `mad`.`pokemon`
;create view settings_area as
select * from `mad`.`settings_area`
;create view settings_device as
select * from `mad`.`settings_device`
;create view trs_status as
select `instance_id`,`device_id`,st_geometryfromwkb(point(st_y(`currentPos`),st_x(`currentPos`))) AS `currentPos`,st_geometryfromwkb(point(st_y(`lastPos`),st_x(`currentPos`))) AS `lastPos`,`routePos`,`routeMax`,`area_id`,`idle`,`rebootCounter`,`lastProtoDateTime`,`lastPogoRestart`,`rebootingOption`,`restartCounter`,`lastPogoReboot`,`globalrebootcount`,`globalrestartcount`,`currentSleepTime`,`last_softban_action`,`last_softban_action_location` from `mad`.`trs_status`
```

local.json
```json
"schemas": [
  {
    "type": "mad",
    "host": "$MAD_DB_IP",
    "port": 3306,
    "username": "$MAD_DB_USERNAME",
    "password": "$MAD_DB_PASSWORD",
    "database": "$MAD_DB_DATABASE",
    "charset": "utf8mb4",
    "hasAltQuests": false,
    "useFor": [
      "gym",
      "pokemon",
      "pokestop",
      "scanCell",
      "spawnpoint",
      "weather",
      "nest"
    ]
  },
  {
    "type": "mad",
    "host": "$MAD_DB_IP",
    "port": 3306,
    "username": "$MAD_DB_USERNAME",
    "password": "$MAD_DB_PASSWORD",
    "database": "maddummy",
    "charset": "utf8mb4",
    "hasAltQuests": false,
    "useFor": [
      "device"
    ]
  }
]
```

TurtIeSocks:
`type`, `hasAltQuests`, and `charset` are all deprecated and no longer used.

crhbetz:
> I believe @crhbetz is also running the async branch, can you confirm this is an issue for you as well?

In case it's still relevant: yes, I can confirm this. I'm scanning in southern Germany and the scanners are displayed in Somalia on ReactMap :-)

fosJoddie:
This us indeed wrong in async branch. Please hold on this PR for a fix in async

TurtIeSocks:
Going to close since this will be addressed in MAD

PR #576, opened 2022-10-06
Extract Config Pages

### LINKED ISSUE #574: [Feature Request] Extract config for MOTD, donor and login page to separate file
local.json would be more readable if we extract that json nodes to a separate file

TurtIeSocks:
- Allow `messageOfTheDay.json`, `loginPage.json`, and `donationPage.json` in the config folder
- Works with multiDomains by adding a folder and named jsons instead: `server/src/configs/loginPage/subDomain_1.json`
`server/src/configs/loginPage/subDomain_2.json`

Resolves #574

### DIFF
--- a/server/src/services/config.js
+++ b/server/src/services/config.js
@@ -106,11 +106,12 @@ const checkExtraJsons = (fileName, domain = '') => {
     domain &&
     fs.existsSync(resolve(`${__dirname}/../configs/${fileName}/${domain}.json`))
   ) {
-    const domainJson = JSON.parse(
-      fs.readFileSync(
-        resolve(__dirname, `../configs/${fileName}/${domain}.json`),
-      ),
-    ) || {}
+    const domainJson =
+      JSON.parse(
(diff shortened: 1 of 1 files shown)


TurtIeSocks:
@bbdoc would you be able to test this for multi domains? I vaguely recall you complaining about config length once upon a time.

bbdoc:
Will do... config length is one thing, but I also really like this as it makes it very easy to disable a motd for a subdomain and reenable it later on by simply changing the filename ;-)

Is there a way to create a json for a message that would be valid for all domains or does this one remain in the global config ?

TurtIeSocks:
It layers like this:
1. `server/src/configs/local.json`
2. `server/src/configs/messageOfTheDay.json`
3. `server/src/configs/messageOfTheDay/domain_1.json`

So you could put everything you want shared across your multi domains into `#2`, then domain specifics in `#3`. It only does a shallow merge though, it's not merging the `components` array for example, just overwriting entirely if present.

PR #590, opened 2022-10-14
Scan Next/Zone Discord Logging

TurtIeSocks:
image

**Important**
- REQUIRES ADJUSTMENTS IF YOU USE A CUSTOM STRATEGY, see the changes to the files in `server/src/strategies` 

**Other notes**
- Version bump
- Add thumbnailUrl to discord strategy object
- pass full req.user to scanner api
- Initialize strategies in a separate file for dynamically collecting all clients
- Rudimentary cache for storing # of requests for users in a given day
- Log to the same channel as authentication reqs
- Remove unused accessToken arg
- Save logChanneld in the class instance
- Remove user info from frontend, collect from request instead

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "reactmap",
-  "version": "1.6.2",
+  "version": "1.6.3",
   "description": "React based frontend map.",
   "main": "ReactMap.js",
   "author": "TurtIeSocks <58572875+TurtIeSocks@users.noreply.github.com>",
--- a/server/src/configs/default.json
+++ b/server/src/configs/default.json
@@ -559,6 +559,8 @@
         "type": "discord",
         "enabled": false,
         "logChannelId": "",
+        "scanNextLogChannelId": "",
+        "scanZoneLogChannelId": "",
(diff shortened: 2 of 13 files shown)


kamieniarz:
1. When that request count resets (if it does)?
2. Can we have separate log channel for that?

AppIe314:
I've just skimmed through the PR. Should the coords payload be trimmed if it exceeds something like 10 coords? IIRC scanZone has up to 300+ requests (which is a coordinate?) so the Discord message could be very long.

TurtIeSocks:
@kamieniarz 
Resets once a day or whenever the map server is reset. It's a rudimentary implementation that's mostly meant for observing very obvious signs of abuse. At this point, just putting off any significant logging implementation until a rewrite. 

...separate channels are quite annoying due to custom/multiple auth strategies.

@AppIe314 
Coords are trimmed after 25:
https://github.com/WatWowMap/ReactMap/blob/2f25413c5e9b42e004dfcc06be680d9d4b6935b5/server/src/services/api/scannerApi.js#L164

Issue #617, opened 2022-12-06
Arearestrictions exceeds embed limits causing crash.


jepke:
When there is a long list of matched areas and discord logging is enabled. It exceeds the max lenght of 1024 on the discord embed resulting in a crash of Reactmap upon login.

```
/home/node/node_modules/discord.js/src/rest/RequestHandler.js:154
      throw new DiscordAPIError(request.path, data, request.method, res.status);
            ^
DiscordAPIError: Invalid Form Body
embed.fields[15].value: Must be 1024 or fewer in length.
embeds[0].fields[15].value: Must be 1024 or fewer in length.
    at RequestHandler.execute (/home/node/node_modules/discord.js/src/rest/RequestHandler.js:154:13)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async RequestHandler.push (/home/node/node_modules/discord.js/src/rest/RequestHandler.js:39:14) {
  method: 'post',
  path: '/channels/#############/messages',
  code: 50035,
  httpStatus: 400
}
```

jepke:
#618 as fix for this

PR #618, opened 2022-12-06
Discord max field size 1024 exceeded fix

jepke:
There are more elegant ways to fix this but its a fix.

### DIFF
--- a/server/src/services/DiscordClient.js
+++ b/server/src/services/DiscordClient.js
@@ -127,8 +127,11 @@ module.exports = class DiscordMapClient {
       if (perms.areaRestrictions.length) {
         perms.areaRestrictions = [...new Set(perms.areaRestrictions)]
       }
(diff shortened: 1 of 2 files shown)


jepke:
nvm this breaks even more... WIP

TurtIeSocks:
Instead of setting this in the discord client, which sets the permissions for the user across the entire site, you'll want to limit it in the embed building function:
https://github.com/WatWowMap/ReactMap/blob/main/server/src/services/logUserAuth.js

TurtIeSocks:
I'm doing something similar here for scanNext/scanZone logging you can reference:
https://github.com/WatWowMap/ReactMap/blob/38c25266a4cc9e15d6ff2a5c6d701352a3895885/server/src/services/api/scannerApi.js#L167

jepke:
fix for #617

TurtIeSocks (approved):

PR #620, opened 2022-12-12
Update Pokestop.js

spammer23:
Remove duplicated 'AS grunt_type'


TurtIeSocks (approved):
I don't have a MAD database anymore that I can test with, I assume you've confirmed that this is working correctly?

PR #622, opened 2022-12-15
Size Proto Support

TurtIeSocks:
- Adds a few translation keys
- Moves all "old size" reference to height
- New size reference is the size proto
- Maintains backwards compatibility
- Filter size globally and per pokemon
- Adds map marker indicator to view the size of the pokemon
- Requires updates from those using custom pokemon markers
- Also adds an indicator in the popup
- Version bump

Requires https://github.com/RealDeviceMap/RealDeviceMap/pull/456


server/src/graphql/scannerTypes.js
```diff
@@ -134,6 +134,7 @@ module.exports = gql`
     display_pokemon_id: Int
     ditto_form: Int
     weight: Float
+    height: Float
```
Fabio1988, line 137: height is Float, size Int..?!

TurtIeSocks: Maintains backwards compatibility for now, since `size` for everyone not using the latest rdm is still going to be a float ;)

PR #623, opened 2022-12-19
add poracle size support

TurtIeSocks:
- add translation key for `slider_size`
- bump required poracle version
- add size selection components
- add custom mark and tick support to sliders


src/services/Poracle.js
```diff
@@ -560,6 +562,16 @@ export default class Poracle {
               }-${item.max_sta}${item.clean ? ` | ${t('clean')} ` : ''}${
                 item.distance ? ` | d${item.distance}` : ''
               }`
+        }${
+          item.size > 1 || item.max_size < 5
+            ? `${t('size', 'Size')}:${
```
kamieniarz, line 567: ```            ? ` | ${t('size', 'Size')}:${```
| is missing ;)

PR #626, opened 2022-12-21
profile page fixes

TurtIeSocks:
- Fix misnamed i18n key
- Add key for scanner_subtitle
- Show missing perms for non-logged in users that don't have access to them
- Fix scanner perm display
- Properly account for perms that have empty arrays


jepke:
`Properly account for perms that have empty arrays`
Just perms or does it go for empty arearestrictions as well

Fabio1988 (approved):

TurtIeSocks:
> `Properly account for perms that have empty arrays` Just perms or does it go for empty arearestrictions as well

Screenshot 2022-12-22 at 12 37 40 PM
Now it just grays them out and doesn't count them towards the total perms that they might have in the tutorial if they have 0, since previously an empty array was considered truthy. Though now I'm thinking that maybe I should add some logic to it to just not show them at all if there aren't any area restrictions/scanner tools enabled/webhooks added...

jepke:
In my case users are required to pick their areas they want to see. If they are dumb and do not read and dont pick any they currently just get all available areas. I'd rather have them get nothing. My current work around is to have discord guild id in arearestrictions with a tiny dummy fence in the middle of a lake.

Ideally if they dont have an area or other specific mandatory perms set a popup will direct them to either discord faq channel or something but that might get complex real quick with all the flexibility rm has.

TurtIeSocks:
I think latest commit vaguely fixes what you're talking about @jepke, if you want to try it before I merge tonight.

PR #632, opened 2023-01-07
Include Kacleon Stops in Event Stops

ccev:
This should change the event stop query so that it includes all display types >= 7. Right now this only includes the Coin and Kacleon Stops. This is assuming that future event stops work the same way.

Untested

### DIFF
--- a/server/src/models/Pokestop.js
+++ b/server/src/models/Pokestop.js
@@ -519,7 +519,7 @@ module.exports = class Pokestop extends Model {
         if (onlyEventStops && pokestopPerms) {
           stops.orWhere((event) => {
             event
-              .where('display_type', 7)
+              .where('display_type', '>=', 7)
               .andWhere('character', 0)
               .andWhere(
                 multiInvasionMs ? 'expiration_ms' : 'expiration',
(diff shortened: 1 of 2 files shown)


kamieniarz:
Doesn't work for me :(

ChaosEmpire:
Working fine...like the Golden Copins..so it display orange Stops (not kecleon)

nileplumb:
Not working here really sadly

Pogodenhelder:
tried it as well, and it didnt work...

lenisko:
Can't check now, but it's most likely connected to filtering https://github.com/WatWowMap/ReactMap/blob/develop/server/src/models/Pokestop.js#L588-L597

Pogodenhelder:
Schermafbeelding 2023-01-07 om 20 08 35
this is how it looks on the map.
The golden coin or how it looks like and the timer most likely gives it away.. but im not sure.

nileplumb:
Ideally it should start looking at uicons for_i8 stops I already populated my repos with them

nileplumb:
F9A2F326-DA9E-4CA6-9B63-210E5D5EEA72

Ninjasoturi:
https://github.com/ninjasoturi/ReactMap/tree/kecleo
Dirty fix until some actual programmer can come up with a better one

TurtIeSocks (approved):
Thanks! Made some fixes, working now.

PR #634, opened 2023-01-08
splits event stop permissions

### LINKED ISSUE #633: Event Stops do not respect incident perms
Event stops are seen for everyone.

auth Config:
- Without any perm restrictions:
  - map
  - gyms
  - pokestops
  - raids
- all other perms restricted with donor role id

PS: I then tried also to restrict pokestops with donor id, but it seems it's still visible for non registered user. My map is visible for everyone visiting, no login needed. I only restrict premium features like pokemon, invasions, quests, etc

TurtIeSocks:
**Breaking-ish changes**
- Adds a new perm for event stops
- Version bump
- Forced logout for your users
- New perm images and subtitle
- Routing fixes for vite usage
- Fixes event tooltip 

Resolves #633


kamieniarz:
I ran some tests with accounts with different roles and permissions and all seems fine to me. Thanks!

kamieniarz:
Shit, actually I get this error when I enabled all pokestops:
`Invariant Violation: 34 (see https://github.com/apollographql/invariant-packages)`
One of my users also reported me that

TurtIeSocks:
Not sure what that could mean and not seeing it myself when I enable all pokestops.

kamieniarz:
In my case I received that error right after I enabled all pokestops - I had to clear cookies and restore my settings and I can't reproduce that error now.

TurtIeSocks:
Okay managed to force a similar error, though a different combinations of switches. Pushed the fix.

kamieniarz:
Works now, thanks!

Fabio1988:
Works :)

Fabio1988 (approved):

PR #653, opened 2023-02-06
Attempt to add scout

lenisko:
Custom worker to process scans in desired location.

`POST /` - with `[[1,2],[3,4]]` as input data in body. Resturns `{"queue":3}`
`GET /queue` - resturns `{"queue":3}`


lenisko:
not working, I'll try later

PR #658, opened 2023-02-13
Fixed SQL error

Ninjasoturi:


TurtIeSocks (commented):
Thanks for the PR, but what is the reason for the changes? Discord and Telegram IDs are stored as strings but the user ID is stored is an int and I have verified that the existing solution is working.

Ninjasoturi:
Damn, how tired have I been when I was looking at this 😅 I was trying to clear sessions using the session id and that resulted in an error. And I didn't even notice that my "fix" didn't clear the session. You're right, the existing solution is working 👍 Closing this

PR #666, opened 2023-03-14
Update translation de.json

ReuschelCGN:


public/base-locales/de.json
```diff
@@ -281,6 +281,7 @@
   "show_ex_badge": "EX Abzeichen anzeigen",
   "search": "Suche",
   "global_search_quests": "Füge den Name der Belohnung ein...",
+  "global_search_lures": "Füge den Name des Lockmodule ein...",
```
Fabio1988, line 284: Namen des Lockmoduls ?

Fabio1988: Oh also in the other translations it uses "den Name des" 😵‍💫

ReuschelCGN: changed all

Fabio1988 (approved):

PR #678, opened 2023-03-30
Update Locale for Event Stops to Kecleon

darthbutcher:
As Niantic has moved Golden Stops to become Sparkly Lures, currently the Event Stop is only used for Kecleon.

This will allow a locale solution to provide users with the information about where to easily filter for Kecleon, but leaves the codebase in tact for event_stops so as to allow us to easily revert this change in the future if Niantic adds more options.


TurtIeSocks:
@ReuschelCGN @Fabio1988, going to merge this, but you may want to PR german translations in another PR.

TurtIeSocks (approved):

darthbutcher:
The German name for Kecleon is Kecleon. I looked it up before I made this PR.

ReuschelCGN:
yes but eventstops will come back, when pokemon contests will find its way into game

TurtIeSocks:
This is just a temporary patch since I don't have time to address it right now. A future PR will remove it and add selecting of each individual stop display type. @ReuschelCGN

PR #690, opened 2023-04-15
bLaZiNgLy Fast Pokemon

TurtIeSocks:
- Better in memory implementation of the Pokemon table
- At least 5-10x faster "query" times for PVP filtering
- Requires https://github.com/UnownHash/Golbat/pull/105
- add `pokemon` property to your golbat schema with the value of `http://{golbat_host}}:{{golbat_port}}`


server/src/models/Pokemon.js
```diff
@@ -415,22 +519,37 @@ module.exports = class Pokemon extends Model {
...
+      }
+      if (mem && typeof query === 'string') {
+        fs.writeFileSync(
+          resolve(__dirname, './queries', `${Date.now()}.json`),
+          query,
+        )
+      } else {
+        fs.writeFileSync(
+          resolve(__dirname, './queries', `${Date.now()}.txt`),
+          query.toKnexQuery().toString(),
```
na-ji, line 535: When querying available, the query is undefined, so you might want to check that we at least have a query.

TurtIeSocks:
**Breaking changes for those on this branch**

As of latest commit, your config schemas should look like this now:
```json
    "schemas": [
      {
        "host": "127.0.0.1",
        "port": 3306,
        "username": "golbat_username",
        "password": "golbat_password",
        "database": "golbat",
        "useFor": [
          "scanCell",
          "spawnpoint",
          "gym",
          "pokestop",
          "weather"
        ]
      },
      {
        "type": "golbat",
        "endpoint": "http://127.0.0.1:9001",
        "secret": "golbat_bearer_token",
        "useFor": [
          "pokemon"
        ]
      },
    ]
```

This cleans up the schema a bit, separation of concerns, etc etc. also leaves it open to query both the database and the endpoint, with perhaps some logic in the future to be able to smartly switch between the two if Golbat was recently restarted and doesn't have its in-memory pokemon filled up yet.

PR #722, opened 2023-06-11
show cords in gym / stop / mon / portal popup

ReuschelCGN:


TurtIeSocks (changes requested):
Thank you for the PR!

Since all of them are the exact same, perhaps we could make a shared component in the common folder, then pass the lat and lon as props :)

PR #726, opened 2023-06-15
Move Coords to Client Side Settings

plinytheelder:
Don't default everything to a spoofer map. Move Coords to client side settings.

### DIFF
--- a/public/base-locales/en.json
+++ b/public/base-locales/en.json
@@ -590,8 +590,12 @@
   "delete": "Delete",
   "name": "Name",
   "only_confirmed": "Only Confirmed",
-  "rocket_pokemon": "Rocket Pokemon",
+  "rocket_pokemon": "Rocket Pokémon",
   "decoy": "Decoy",
   "s2_cell_limit_0": "You attempted to generate more than 20,000 cells ({{variable_0}})",
-  "show_all_pvp_ranks": "Show All PVP Ranks"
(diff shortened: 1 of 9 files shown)


TurtIeSocks (changes requested):
Dayummm, look at you! Only feedback that comes to mind at first glance is please add english translation keys into the base locales file. There are some "automatic" translations done when keys are missing, so it appears fine when browsing in English, but it's important to maintain the keys in the json for other locales that want to add them. `enablePokemonPopupCoords` => `enable_pokemon_popup_coords` in the `en.json` file.

Great job, thanks for the PR!

plinytheelder:
no problemo! let me know if I did it right this time.

TurtIeSocks (approved):

Fabio1988:
Could it be that this is completly disabling the possibility to disable that feature for the admin?

For me e.g I would nerver give user the possibility to show the coords

TurtIeSocks:
Someone can PR something like this:
https://github.com/WatWowMap/ReactMap/blob/main/server/src/services/ui/clientOptions.js#L96-L98

ReuschelCGN:
like this i build this morning?
https://github.com/ReuschelCGN/ReactMap/blob/666ce7194a090f1e8144f746efa26eac75ca4155/server/src/services/ui/clientOptions.js#L106

PR #727, opened 2023-06-16
bring back popupcoords under admin control

ReuschelCGN:
- fix Lint
- update german translations
- make userslider for coords in popup shareable in config (for gyms/stops/mons/portals)
  (default: false)
- fix Lint another time (i hate that)


server/src/services/checkForUpdates.js
```diff
@@ -43,7 +43,7 @@ try {
               : stdout2.trim()
 
           exec(
-            `git ls-remote https://github.com/WatWowMap/ReactMap/ refs/heads/${branch}`,
+            `git ls-remote https://github.com/ReuschelCGN/ReactMap/ refs/heads/${branch}`,
```
clburlison, line 46: This shouldn't be here 
```suggestion
            `git ls-remote https://github.com/WatWowMap/ReactMap/ refs/heads/${branch}`,
```

server/src/services/ui/clientOptions.js
```diff
@@ -1,6 +1,10 @@
 const {
   clientSideOptions,
   map: { enableMapJsFilter },
+  map: { enablePokemonPopupCoordsSelector },
+  map: { enableGymPopupCoordsSelector },
+  map: { enablePokestopPopupCoordsSelector },
+  map: { enablePortalPopupCoordsSelector },
```
TurtIeSocks, line 7: One small thing here, instead of importing them like this, please consolidate it to be like this:
```js
const {
  clientSideOptions,
  map: { enableMapJsFilter, enablePokemonPopupCoordsSelector, ...the rest },
  api: {
    pvp: { levels },
  },
} = require('../config')
```
replacing `the rest` with the other imports.

TurtIeSocks (changes requested):
Thanks for the quick turn around. Apologies, I should have checked with you before merging in Pliny's PR. Left a comment on one small thing. You mentioned you didn't lint but the linting commands didn't make any changes so I guess it's good to go? Either way, in the future, if you weren't aware, you can run `yarn prettier:fix && yarn eslint:fix` to save some time.

TurtIeSocks (approved):

PR #729, opened 2023-06-22
User feedback on Discord guild blocking & no map access

clburlison:
# Problem
Discord guild blocking has been broken for a while maybe since the MapJS port? It was never a major issue but the current implementation definitely doesn't look correct. Currently when a user attempts to sign-in and is included in a blocked Discord Guild they will return to the login page and no discord alert is ever sent.

<details>
<summary>
Server side error before this PR
</summary>

    ℹ 2023-06-22 00:59:30 [EVENT] Added Spinda Key: 327-0 to masterfile. (pokestops)
    Server is now listening at http://0.0.0.0:8060
    Server is now listening at http://0.0.0.0:8060
    ⚠ 2023-06-22 00:59:41 [DISCORD] Failed to get perms for user 130865078219046915 TypeError: Cannot read properties of undefined (reading 'find')
        at DiscordClient.getPerms (/Users/clayton/src/scanner/ReactMap/server/src/services/DiscordClient.js:133:42)
        at DiscordClient.authHandler (/Users/clayton/src/scanner/ReactMap/server/src/services/DiscordClient.js:240:27)
        at Strategy._verify (/Users/clayton/src/scanner/ReactMap/server/src/services/DiscordClient.js:334:27)
        at /Users/clayton/src/scanner/ReactMap/node_modules/passport-oauth2/lib/strategy.js:198:24
        at /Users/clayton/src/scanner/ReactMap/node_modules/passport-discord/lib/strategy.js:99:24
        at /Users/clayton/src/scanner/ReactMap/node_modules/passport-discord/lib/strategy.js:115:13
        at passBackControl (/Users/clayton/src/scanner/ReactMap/node_modules/oauth/lib/oauth2.js:134:9)
        at IncomingMessage.<anonymous> (/Users/clayton/src/scanner/ReactMap/node_modules/oauth/lib/oauth2.js:157:7)
        at IncomingMessage.emit (node:events:525:35)
        at IncomingMessage.emit (node:domain:489:12)
        at endReadableNT (node:internal/streams/readable:1359:12)
        at process.processTicksAndRejections (node:internal/process/task_queues:82:21)
    ⚠ 2023-06-22 00:59:41 [DISCORD] 130865078219046915 Not authorized to access map

</details>

# The fix

We now include a nice UI for the end user telling them that:
A. They were blocked on purpose
B. Which discord server(s) are on the block list
C. How to get more information by joining discord

This also now properly fires a discord alert to the audit log. Also if a RM user doesn't have the `discordLink` option enabled the Join button isn't shown.

PS - I'm aware the Discord Join button is not the same as the Login page but this does align with the 404 error page. It looks fine to me 🤞  

### Screenshots

Screenshot 2023-06-21 at 7 52 15 PM
Screenshot 2023-06-21 at 7 53 57 PM
Screenshot 2023-06-21 at 7 58 06 PM
 
## Update 
After talking in discord I figured we can add the UI feedback when users don't have map access.
 
Screenshot 2023-06-22 at 10 13 48 PM

### DIFF
--- a/server/src/services/DiscordClient.js
+++ b/server/src/services/DiscordClient.js
@@ -253,11 +253,11 @@ module.exports = class DiscordClient {
       this.sendMessage(embed)
 
       if (user.perms.blocked) {
-        const guildArray = Array.from(user.perms.blockedGuildNames)
+        const guildArray = [...user.perms.blockedGuildNames]
         const lastGuild = guildArray.pop()
         const guildString =
           guildArray.length === 1
-            ? `${guildArray.join(', ')} & ${lastGuild} `
+            ? `${guildArray.join(', ')} & ${lastGuild}`
             : lastGuild
         return done(null, false, {
           message: guildString,
--- a/server/src/services/DiscordClient.js
+++ b/server/src/services/DiscordClient.js
@@ -260,7 +260,7 @@ module.exports = class DiscordClient {
             ? guildArray.join(', ') + ' & ' + lastGuild
             : lastGuild
         return done(null, false, {
-          message: `You have been blocked for being a member of ${guildString}.`,
+          message: guildString,
         })
       }
       if (user) {
--- a/server/src/services/DiscordClient.js
+++ b/server/src/services/DiscordClient.js
@@ -259,9 +259,10 @@ module.exports = class DiscordClient {
           guildArray.length === 1
             ? `${guildArray.join(', ')} & ${lastGuild}`
             : lastGuild
-        return done(null, false, {
-          message: guildString,
-        })
+        return done(null, false, { blockedGuilds: guildString })
+      }
+      if (user.perms.map === false) {
+        return done(null, false, { message: 'access_denied' })
       }
       if (user) {
         delete user.guilds
--- a/server/src/services/DiscordClient.js
+++ b/server/src/services/DiscordClient.js
@@ -139,9 +139,6 @@ module.exports = class DiscordClient {
               : new Set([currentGuildName])
           }
         }
-        if (perms.blocked === true) {
-          return perms
-        }
         await Promise.all(
           this.strategy.allowedGuilds.map(async (guildId) => {
             if (guilds.includes(guildId)) {
@@ -253,7 +250,7 @@ module.exports = class DiscordClient {
       this.sendMessage(embed)
 
       if (user.perms.blocked) {
-        const guildArray = [...user.perms.blockedGuildNames]
+        const guildArray = user.perms.blockedGuildNames
         const lastGuild = guildArray.pop()
         const guildString =
           guildArray.length === 1
--- a/server/src/services/DiscordClient.js
+++ b/server/src/services/DiscordClient.js
@@ -114,7 +114,6 @@ module.exports = class DiscordClient {
     }
 
     try {
-      const { guildsFull } = user
       const guilds = user.guilds.map((guild) => guild.id)
       if (this.strategy.allowedUsers.includes(user.id)) {
         Object.keys(this.perms).forEach((key) => (perms[key] = true))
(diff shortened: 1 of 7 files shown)


src/components/layout/auth/Blocked.jsx
```diff
@@ -0,0 +1,86 @@
...
+import { useTranslation } from 'react-i18next'
+import { Grid, Typography, Button } from '@material-ui/core'
+
+function decodeQueryParam(p) {
+  return decodeURIComponent(p.replace(/\+/g, ' '))
+}
+
+export default function Blocked({ serverSettings }) {
+  const { t } = useTranslation()
+  const params = window.location.href.split('/').pop()
```
TurtIeSocks, line 12: I think ReactRouter might have a nice hook for getting params.

server/src/services/DiscordClient.js
```diff
@@ -114,7 +114,6 @@ module.exports = class DiscordClient {
     }
 
     try {
-      const { guildsFull } = user
```
TurtIeSocks, line 117: Not sure how this one escaped me, but good bug fix!

server/src/services/DiscordClient.js
```diff
@@ -127,13 +126,22 @@ module.exports = class DiscordClient {
...
+            const currentGuildName = guildsFull.find(
+              (x) => x.id === guildId,
+            ).name
+            perms.blockedGuildNames = perms.blockedGuildNames
+              ? perms.blockedGuildNames.add(currentGuildName)
+              : new Set([currentGuildName])
           }
         }
+        if (perms.blocked === true) {
+          return perms
```
TurtIeSocks, line 143:
```suggestion
          perms.blockedGuildNames = Array.from(perms.blockedGuildNames)
          return perms
```
Lets turn the set into an array here to be consistent with the other perms. Then we can do just do normal array operations in the other places you're using `Array.from()`.

src/components/layout/auth/Blocked.jsx
```diff
@@ -0,0 +1,86 @@
...
+            item
+            xs={t('join')}
+            sm={3}
+            style={{ textAlign: 'center', marginTop: 20 }}
+          >
+            <Button
+              variant="outlined"
+              color="secondary"
+              onClick={() =>
+                (window.location = serverSettings.config.map.discordInvite)
```
TurtIeSocks, line 76: Hmm maybe this should open in a new window or did you intentionally want to bring them away from the blocked page?

TurtIeSocks (changes requested):
Great fix and QoL changes! I'll test the UI a bit later but I left some small comments. Thank you!

src/components/layout/auth/Blocked.jsx
clburlison: It does and I've used it before however I copy/pasted this from the existing code base for consistency. 😂 I can switch it tomorrow.

src/components/layout/auth/Blocked.jsx
clburlison: Was copy/pasting. New window is likely a better action.

server/src/services/DiscordClient.js
clburlison: For reference you're actually using a Set for the webhook & scanner perms. https://github.com/WatWowMap/ReactMap/pull/729/files#diff-fcad83d8a0f164f00fdbb33b303964ac83270bada9a2e90e1078aec2b46f13a5R109-R113

That that said you are using spread operators instead of `Array.from()` in the rest of the codebase so I have switched to that.

TurtIeSocks: Sorry for the confusion, I am using sets within this function but I turn them all into arrays before returning https://github.com/WatWowMap/ReactMap/blob/d812c48b255a1d83102c26e458edf00a0ceedf35/server/src/services/DiscordClient.js#L197, is what I was referencing.

clburlison: ew. I didn't see that.

clburlison: Resolved properly now

src/components/layout/auth/Blocked.jsx
```diff
@@ -0,0 +1,90 @@
...
+      <Grid
+        container
+        item
+        justifyContent="center"
+        alignItems="center"
+        style={{ marginTop: 20, paddingTop: 20, marginBottom: 20 }}
+      >
+        <Grid
+          item
+          xs={serverSettings.config.map.discordInvite ? t('go_back') : 10}
```
TurtIeSocks, line 54: Need to replace `t('go_back')` with a number.

src/components/layout/auth/Blocked.jsx
```diff
@@ -0,0 +1,90 @@
...
+            color="secondary"
+            onClick={() => (window.location = window.location.origin)}
+          >
+            {t('go_back')}
+          </Button>
+        </Grid>
+        {serverSettings.config.map.discordInvite && (
+          <Grid
+            item
+            xs={t('join')}
```
TurtIeSocks, line 72: Same here.

TurtIeSocks (changes requested):
Tested and works great! Just two small corrections needed on the frontend portion.

TurtIeSocks (approved):
Thank you!

PR #734, opened 2023-07-01
uicons try/catch in browser

TurtIeSocks:
- Adds a backup object in case index fetching fails 
- Adds try/catch blocks to all uicons methods in the client, since this seems to be the most common crash and won't go away until github stops sucking


matt-yml:
I did enjoy github's fail messaging. It's somewhat suggestive.

image
image

PR #736, opened 2023-07-06
feat: individual event filtering

TurtIeSocks:
- split event stops into separate filters
- new collapse menu when enabling `Event Stops` in the drawer
- poracle support for `showcase`
- waiting for better backend support to fill in popup info before merge

Screenshot 2023-07-06 at 1 34 11 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "reactmap",
-  "version": "1.16.1",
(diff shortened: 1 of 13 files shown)


TurtIeSocks:
Nvm, merging this now since it looks like it'll be a while before the backends are ready. I'll open a separate PR when they are.

PR #751, opened 2023-07-11
feat: interactive marker opacity

### LINKED ISSUE #683: Visual indication of a despawning mon/invasion
Maybe can make a mon semi-transparent if it is going to despawn within a minute (if timer is verified)? :)

TurtIeSocks:
- Sets opacity of the marker based on the despawn time
- Active for pokemon, invasions, and raids
- Follows these rules:
  - < 1 min = 25%
  - < 5min = 50%
  - < 10min = 75%
  - More than 10min = 100%
- This does not interact with the marker memoization equality checks for performance reasons. In order to update opacity, you'll have to move the marker off screen and back again to rerender it

Resolves #683

### DIFF
--- a/src/components/markers/gym.jsx
+++ b/src/components/markers/gym.jsx
@@ -1,6 +1,7 @@
 import React from 'react'
 import { renderToString } from 'react-dom/server'
 import L from 'leaflet'
+import getOpacity from '@services/functions/getOpacity'
 
 const getBadgeColor = (raidLevel) => {
   switch (raidLevel) {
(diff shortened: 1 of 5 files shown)


kamieniarz:
Please make it optional for users but switched off by default

plinytheelder:
my users and I agree with @kamieniarz. this should be optional.

TurtIeSocks:
All of my users have loved it so far. PRs will be accepted though 👍

plinytheelder:
can look into it. a variable should be easy.

but since I'm leaving for work, setting all these variables to 1 is an ez bandaid fix.

https://github.com/WatWowMap/ReactMap/blob/develop/src/services/functions/getOpacity.js

PR #752, opened 2023-07-11
feat: scan on demand cooldowns

### LINKED ISSUE #586: scanNext cooldown for users
image
After a user uses this, having a cooldown would be needed I think for widespread use. Maybe there is one and i'm missing it?

TurtIeSocks:
- Adds configurable cooldown periods between each scanNext and scanZone request
- The cooldown time is multiplied by the number of requests sent by the user

Resolves #586


src/components/layout/dialogs/scanner/ScanOnDemand.jsx
```diff
@@ -69,6 +72,12 @@ export default function ScanOnDemand({
     if (scanMode === 'sendCoords') {
       demandScan()
       setScanMode('loading')
+      const timer = mode === 'scanNext' ? scanNextCooldown : scanZoneCooldown
+      useStatic.setState({
+        scannerCooldown:
+          (typeof timer === 'number' ? timer : 0) *
+          (mode === 'scanNext' ? scanCoords.length : scanZoneSize),
```
kamieniarz, line 79: I think `scanCoords.length` for both types would be better. It would be easier to manage cooldown especially when scanzone size 1 = 7 coords, size 2 = 19, size 3 = 37 and so on

kamieniarz:
I love the idea and it works pretty well but I have some thoughts on that:

- config cooldown could be always multiplied by scanCoords.length (already pointed that out above)
- cooldown could be binded with user cause of:
   - countdown works only when scanzone/scannext popup is open
   - refreshing map resets the cooldown

PR #761, opened 2023-07-12
Bring back map markers when using scanZone

kamieniarz:
In https://github.com/WatWowMap/ReactMap/pull/580 map markers were disabled on scanZone feature due to performance issues. I think we're past these issues so we can bring markers back. 
I ran some tests locally on desktop and mobile and I hardly can see any difference when using map with markers or map with markers and scanZone on top

Btw I hope that's the only change that needs to be done ;)


TurtIeSocks (approved):
I have concerns about usability in some scenarios but I can't seem to replicate the issue that was the original reason for adding this line so I'll approve and revisit if the issue pops up again.

PR #763, opened 2023-07-16
Update Scanner local perms + optional Contribute

Petap0w:
Make Contribute menu button optional (active by default)
Update some FR translations
Update Scanner perms in local auth


server/src/configs/default.json
```diff
@@ -605,6 +606,7 @@
       "userCooldownSeconds": 0,
       "scanNextAreaRestriction": [],
       "discordRoles": [],
+      "localStatus": [],
```
TurtIeSocks, line 609: Let's just call this `local`.

src/components/layout/drawer/Actions.jsx
```diff
@@ -157,20 +157,22 @@ export default function DrawerActions() {
...
-        style={{ textDecoration: 'none', color: 'white' }}
-      >
-        <ListItem button>
-          <ListItemIcon>
-            <HeartIcon color="primary" />
-          </ListItemIcon>
-          <ListItemText primary={t('contribute')} />
-        </ListItem>
-      </a>
+      {config.map.enableContribute && (
```
TurtIeSocks, line 160: I've been _very_ lenient on what things are configurable in this project, including, removing any references to this project, but I'm going to put my foot down on this one. There are plenty of users that aren't in the mapping sphere that are able to contribute, and with all of the time and effort I've put into it, I think it's perfectly reasonable to leave a single link for those that are interested to find the source. Even if it only attracts a single person who wants to contribute and make the project better, then it's worth it.

I would be willing to bet that this person who recently starred the repo found it from this link :)
Screenshot 2023-07-18 at 2 45 45 PM

TurtIeSocks (changes requested):
Thanks for the PR! 

Please make the changes I mentioned in the comments. I'll need the `enableContribute` configurable value removed before merge.

src/components/layout/drawer/Actions.jsx
svenberg: I agree that the option to contribute should be available, and I agree with your arguments as well @TurtIeSocks. Perhaps under Menu - Settings is the best place, perhaps not. I'm not sure where else to put it though as it shouldn't be hidden, My small concern is that my community believes they are contributing to me - I have had a few questions about it already. I explained what you are saying as well, and that all suggestions and issues are welcome to put there. Most of the stuff however is resolved within the community because an issue is rarely an issue but instead a user error (or not well enough explained. which could be a contribution). Not sure if it makes any sense what I'm saying 😄

PR #764, opened 2023-07-19
Upgrade to Mui v5

TurtIeSocks:
Upgrade the MUI packages to V5

- Properly themed dark/light modes
- Better color theming
- Move to `sx` prop and away from `useStyles`
- New search modal theme


clburlison:
This is working well over here. Super cosmetic. The new search modal has a few oddities:
- the transition on/off screen is weird 
- light mode doesn't look as great. Not 100% confident but think it's the black shadow on the boxes. Might look better with a tiny bit of padding between the two input boxes? Dark mode looks fine IMO.
IMG_1829

iOS Safari 16

TurtIeSocks:
Fixed in latest, thanks!

PR #769, opened 2023-07-21
Level circle

lenisko:
Add level badge. And that's it :)


src/components/markers/pokemon.jsx
```diff
@@ -31,6 +31,7 @@ export const fancyMarker = (
   pkmn,
   glow,
   ivCircle,
+  levelCircle,
```
TurtIeSocks, line 34: Please add this new arg as the last one for the function, to help avoid breaking any custom marker files people may have.

TurtIeSocks (changes requested):
Looks great! Just one small code change and please add the respective translations as well to `public/base-locales/en.json`. The key is a snake case version of the camelCase config name.

TurtIeSocks (approved):

PR #778, opened 2023-07-24
update german translations de.json

ReuschelCGN:


public/base-locales/de.json
```diff
@@ -603,5 +603,24 @@
...
+  "mi": "mi",
+  "filter_by_areas": "nach Gebieten filtern",
+  "submit_nest_name": "Nestnamen mitteilen",
+  "nest_submission_menu": "Namen aktualisieren",
+  "nest_submissions": "Nest Einreichungen",
+  "nest_submissions_subtitle": "Ermöglicht neue Namen für Nester einzureichen.",
+  "pokemon_opacity": "dynamische Pokemon Durchsichtigkeit",
+  "invasion_opacity": "dynamische Invasion Durchsichtigkeit",
+  "raid_opacity": "dynamische Raid Durchsichtigkeit",
+  "opacity_ten_minutes": "Durchsichtigkeit ab 10 Minuten",
```
Fabio1988, line 619: Opacity is more like "Deckkraft",right? :)

Fabio1988: yeah in the context it's better to use Durchsichtigkeit, I agree on that one :) Discussed with author :)

PR #780, opened 2023-07-27
feat: ai generated locales

TurtIeSocks:
- openai/chatgpt based script to generate missing locales 
- github action to generate missing locales when merged into `main`
- build locales during vite process instead of separately
- move base locales into their own folder instead of in public


TurtIeSocks:
@ReuschelCGN @Fabio1988 comments on the added German translations?

Fabio1988 (approved):

PR #781, opened 2023-07-27
feat: initial route support

TurtIeSocks:
- Only compatible with https://github.com/UnownHash/Golbat/pull/133
- New `routes` permission needs to be added to use
- Be sure to add `route` to the appropriate `useFor` array in your config
- May want to force a user logout for changes to go through
- Based on #780


clburlison:
After sleeping on the implementation. It might be a good idea to add the Google/Waze/Apple Map icon. This is a map after all and I expect there will be some users that need navigation in order to drive to new routes. Urban city sprawl 🦜 🥳

TurtIeSocks:
> After sleeping on the implementation. It might be a good idea to add the Google/Waze/Apple Map icon. This is a map after all and I expect there will be some users that need navigation in order to drive to new routes. Urban city sprawl 🦜 🥳

Added in latest.

locales/en.json
```diff
@@ -626,5 +626,11 @@
   "level_circles": "Level Indicators",
   "min_level_circle": "Minimum Circle Level",
   "mutation_auth_error": "Your request was unsuccessful due to not being logged in",
-  "submitted_by": "Submitted By"
-}
+  "submitted_by": "Submitted By",
+  "reversible": "Reversible",
+  "version": "Verison",
```
clburlison, line 631:
```suggestion
  "version": "Version",
```
Submitted by a user

TurtIeSocks: oops, that's embarrassing. Thanks!

PR #788, opened 2023-07-31
Update Fr and local auth fixes

Petap0w:


src/components/layout/auth/Local.jsx
```diff
@@ -40,9 +40,8 @@ export default function LocalLogin({ href }) {
     e.preventDefault()
     setSubmitted(true)
     const resp = await Fetch.login(user, href)
-
-    if (!resp.ok) {
```
TurtIeSocks, line 44: I don't think we should remove this entirely? In case there are errors other than `invalid_credentials`. Can you add it as an else if after `invalid_credentials`?

TurtIeSocks (changes requested):
Everything looks great minus one question about the client login page. Thanks!

src/components/layout/auth/Local.jsx
Petap0w: The response is always 200 ok and only the url changes when it fails authentication (blocked/message=invalid_credentials is added)
But I guess we can leave an else if with another generic failed auth message

PR #808, opened 2023-08-09
Fix typo in config for showcase

Fabio1988:


TurtIeSocks (approved):
Fascinating 😂 good find, thanks!

PR #809, opened 2023-08-10
Support for DNF filters in Golbat

Mygod:
Featuring a much cleaner interface. Old interface is removed for now but could be added back behind a config option in the near future.

Since I am lazy and I also want to prevent the server potentially doing exponentially more computation than the request length (and thus potentially enabling DoS attacks), legacy filters will no longer support the NOT operation (`!`) and parentheses (`()`) after this commit. The client will have to manually flatten the expression into the DNF.

Goes with https://github.com/UnownHash/Golbat/pull/134.


jfberry:
Given the current proposed api does not check global filters if there is a pokemon match, I think this would need to be modified to add the global filter to every pokemon type specifically matched for (I couldn't see that in the code) to reproduce current behaviour [assuming golbat api logic remains unchanged]

Mygod:
https://github.com/WatWowMap/ReactMap/pull/809/files#diff-166340f7ce2e587916e3005ceae01fedc018d36c94d52a0755a7c36b94f801feR320
🤔

jfberry:
> https://github.com/WatWowMap/ReactMap/pull/809/files#diff-166340f7ce2e587916e3005ceae01fedc018d36c94d52a0755a7c36b94f801feR320
> 🤔

Maybe I have misread this but this seemed to be pushing all the leagues assuming an ‘or’ rather than creating individual filters once per league

if I have misread I apologise it’s hard working from a diff when I don’t know the original code either :-)

thunder123456:
If you have global IV filters set to 1-100 and link global and Advanced filters just set just enable one pokemon the query can exceed the max query result in golbat and not show pokemon that are returned when zoomed in.

TurtIeSocks (approved):
Works great, thanks!

PR #813, opened 2023-08-13
Fix MAD event and invasion support

fosJoddie:
Fixes MAD support for event pokestops and also (I think) invasion filtering.

TLDR; MAD DB speaks datetime columns, and some of these are in UTC. 

Also the model to get available events was out of order.

This should in theory work for any lost soul still running on legacy branch of MAD, but I only tested it on async.

### DIFF
--- a/server/src/models/Pokestop.js
+++ b/server/src/models/Pokestop.js
@@ -157,6 +157,7 @@ module.exports = class Pokestop extends Model {
             'pokestop_incident.character_display AS grunt_type',
             'pokestop_incident.incident_display_type AS display_type',
           ])
+          .whereRaw('incident_expiration > UTC_TIMESTAMP()')
       } else {
         query
           .leftJoin('incident', 'pokestop.id', 'incident.pokestop_id')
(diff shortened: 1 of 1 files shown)


server/src/models/Pokestop.js
```diff
@@ -596,25 +597,26 @@ module.exports = class Pokestop extends Model {
         if (onlyEventStops && eventStopPerms && displayTypes.length) {
           stops.orWhere((event) => {
             if (isMad && !hasMultiInvasions) {
-              event
-                .whereIn('incident_grunt_type', MADE_UP_MAD_INVASIONS)
-                .andWhere('incident_expiration', '>=', this.knex().fn.now())
+              event.where(function() { this.whereIn('incident_grunt_type', MADE_UP_MAD_INVASIONS).orWhere('character_display',0) })
```
TurtIeSocks, line 600: Let's change this to an arrow function to match the rest of the codebase.
```suggestion
              event.where((gruntType) => {
                gruntType.whereIn('incident_grunt_type', MADE_UP_MAD_INVASIONS)
                         .orWhere('character_display',0) 
              })
```

TurtIeSocks (changes requested):
Everything seems to work fine as far as I can tell, just one small code style adjustment. I didn't test the suggestion but I'm fairly certain functionality should be the same. Thanks for the PR!

fosJoddie:
I found another error in the query, I'm going to close this one and join it with the other PR to make my life easier.

PR #814, opened 2023-08-13
Route support for MAD db schema

fosJoddie:
This PR will allow people on MAD backend to display routes in RM

this PR needs to be tested by someone with a Golbat backend, to make sure it ain't breaking stuff there
it also needs to be tested by other MAD users

Fixes MAD support for event pokestops and also (I think) invasion filtering.

TLDR; MAD DB speaks datetime columns, and some of these are in UTC.

Also the model to get available events was out of order.

This should in theory work for any lost soul still running on legacy branch of MAD, but I only tested it on async.

### DIFF
--- a/server/src/models/Route.js
+++ b/server/src/models/Route.js
@@ -46,7 +46,7 @@ class Route extends Model {
     const distanceMeters = isMad ? 'route_distance_meters' : 'distance_meters'
     const endLatitude = isMad ? 'end_poi_latitude' : 'end_lat'
     const endLongitude = isMad ? 'end_poi_longitude' : 'end_lon'
-    //route_distance_meters
+    
     const query = this.query()
       .select(isMad ? GET_MAD_ALL_SELECT : GET_ALL_SELECT)
       .whereBetween(startLatitude, [args.minLat, args.maxLat])
@@ -86,7 +86,7 @@ class Route extends Model {
    * @param {boolean} isMad true if DB schema is MAD
    */
   static async getOne(id, { isMad }) {
-    let result = isMad ? 
+    const result = isMad ? 
     await this.query() .select({
         id: 'route_id',
         name: 'name',
--- a/server/src/models/Route.js
+++ b/server/src/models/Route.js
@@ -79,7 +79,8 @@ class Route extends Model {
   /**
    * Returns the full route after querying it, generally from the Popup
    * @param {number} id
-   * @param {boolean} isMad true if DB schema is MAD
+   * @param {import('..types').DbContext} ctx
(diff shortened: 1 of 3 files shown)


mpw1337:
Works on Mad Database, don't have any problems.

andrewagner86:
works for me too.

server/src/models/Route.js
```diff
@@ -66,8 +84,35 @@ class Route extends Model {
    * Returns the full route after querying it, generally from the Popup
    * @param {number} id
    */
-  static async getOne(id) {
-    const result = await this.query().findById(id)
+  static async getOne(id, { isMad }) {
+    let result = null;
```
TurtIeSocks, line 88: Can we use a ternary here? React dev in me tries to avoid mutations when I can :)

server/src/models/Route.js
```diff
@@ -94,13 +139,24 @@ class Route extends Model {
    * returns route context
    * @returns {{ max_distance: number, max_duration: number }}
    */
-  static async getFilterContext() {
+  static async getFilterContext(source) {
+    const { isMad } = source
```
TurtIeSocks, line 143: Can you move this destructuring into the arguments like the other methods do (or should be doing if I missed one)? And add JSdocs with the DbContext type, as above.

TurtIeSocks (changes requested):
Everything seems to work with Golbat as far as I can tell. Thanks for the PR, just two small adjustments.

server/src/models/Route.js
fosJoddie: I couldn't find any example of jsdocs with DbContext type, but I changed the parameter

server/src/models/Route.js
fosJoddie: I'm struggling to let eslint autoformat my document, but I swapped to ternary as pr requested.

fosJoddie:
Closed my other PR and merged it into this, as the changes are "small" enough imo, and it makes it easier to work on

server/src/models/Route.js
```diff
@@ -92,9 +134,19 @@ class Route extends Model {
 
   /**
    * returns route context
+   * @param {boolean} isMad true if DB schema is MAD
```
TurtIeSocks, line 137: Sorry, one last thing! (well technically two...) 

This is an example of adding the `DbContext` to JSdocs, could you please add it here and the one above :)
Screenshot 2023-08-19 at 9 43 40 AM

Then I think we'll good to go!

TurtIeSocks (approved):
Thank you!

PR #832, opened 2023-08-26
Main query/marker rework

TurtIeSocks:
- State management reworks for all map markers
- Removes the large amount of prop drilling that was occurring
- Performance 📈 probably, at a minimum, less embarrassing to look at


Mygod:
📈📈📈 best pr of my life

PR #844, opened 2023-09-12
Update Poracle Translation DE and new ones

Fabio1988:
fix poracle confirm_copy translation

Unfortunately the wording before was bit misleading. not it should be more clean.
When we click on the copy button we want to copy this profile to another. The text says pls choose a profile to copy to this.

@ReuschelCGN your thoughts?


TurtIeSocks (approved):
Thanks!

PR #847, opened 2023-09-14
feat: switch for hiding pii on channel log

lenisko:
quite obvious but untested :)

### DIFF
--- a/server/src/services/logUserAuth.js
+++ b/server/src/services/logUserAuth.js
@@ -2,6 +2,9 @@
 const { default: fetch } = require('node-fetch')
 const { log, HELPERS } = require('@rm/logger')
 
+// PII fields inside getAuthInfo embed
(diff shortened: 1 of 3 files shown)


server/src/services/logUserAuth.js
```diff
@@ -199,6 +200,12 @@ async function getAuthInfo(req, user, strategy = 'custom') {
       'Not authorized to access map',
     )
   }
+  if (hidePii) {
+    embed.fields = embed.fields.filter(field => {
+      const piiFields = ['Ip Address', 'Geo Lookup', 'Google Map', 'Network Provider'];
```
TurtIeSocks, line 205: Can you declare this as a screaming snake case const at the top of the file?

TurtIeSocks (changes requested):
Thanks for the Pr! Just one small change.

TurtIeSocks (approved):

PR #858, opened 2023-10-03
Add route download

na-ji:
Hi!

In this PR, I added a button to download a route as a GPX file. This simply adds a button on the route popup to trigger the action.
image
Since this is a controversial feature, it's hidden behind a config. By default, it's disabled. To enable it, you need to set `map.misc.enableRouteDownload` to true. 

Thanks.


TurtIeSocks (approved):
Lgtm! I'm going to hold off on merging though until after the next dev => main sync. Thank you!

PR #862, opened 2023-10-09
Fix Routes in MAD

fosJoddie:
Raw query for MAD caused a double comparison ">" syntax in SQL which meant no routes where ever returned

#hacktoberfest


TurtIeSocks (approved):
Thanks for the fix!

PR #877, opened 2023-11-05
Add GeoURI navigation option to defaults

frett:
On an Android device it is possible to select the Tesla app as the default handler for Geo URIs.

This allows tapping on a link from your android device to trigger navigation in your Tesla without needing to go through another app and share the link again from there.


TurtIeSocks (changes requested):
Apologies, I missed this. Changes look fine, but can you target the `develop` branch? Thanks for the PR!

frett:
Updated to target develop :)

TurtIeSocks (approved):

PR #922, opened 2024-01-17
Update Child.jsx

wateredstorm:
This update now verifies the screen size. If the screen is small it properly renders area names without overflowing.
Still allows for the previous behaviour when the screen isn't small.

Tested it myself, all functions seem to work accordingly.


HugoDataAnalyst:
I'm not understanding why chore(release) changed the package to a version down.
I did sync the fork before committing :s

PR #931, opened 2024-01-20
Add timestamp to scanZone/scanNext log

kamieniarz:
No idea if that's the best approach but it works for me


TurtIeSocks:
Thanks for the PR but I don't understand what the added value is when there's already a timestamp added to every message footer?

kamieniarz:
It includes seconds

PR #938, opened 2024-01-27
Scout v2 support

kamieniarz:
Requires Dragonite v1.4.1-testing
Set `apiEndpoint` to http://api:port/scout/v2
Getting queue doesn't work (yet?)

I made it with chatgpt so might be not the best approach. Works for me tho

### DIFF
--- a/server/src/configs/default.json
+++ b/server/src/configs/default.json
@@ -595,7 +595,7 @@
   },
   "scanner": {
     "backendConfig": {
-      "platform": "rdm/mad",
+      "platform": "rdm/mad/custom",
       "apiEndpoint": "http://ip:port/api/",
       "routes": true,
       "showcases": true,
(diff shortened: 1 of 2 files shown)


lenisko:
It looks like a breaking change. Add a new type `dragonite` and operate on it instead.

https://github.com/WatWowMap/ReactMap/blob/develop/server/src/services/api/scannerApi.js

Search for `custom` in there, and mostly copy those sections and add a check for `dragonite` type. After that make a changes. You will find queue in there as well.

kamieniarz:
Yeah I agree too. My approach was just a quick one and tbh I have no time to do any adjustments till the weekend so if anyone want's to take over feel free to do so in new PR

kamieniarz:
I messed up sth with github so I opened new PR https://github.com/WatWowMap/ReactMap/pull/942

PR #942, opened 2024-01-29
Scout v2 + its different usage for scanZone/scanNext

kamieniarz:
Changes:

- New platform: `dragonite`. It fully supports scout v2, `custom` stays unchanged. There's no need to change scout `apiEndpoint` if you switch from custom to dragonite
- Using `dragonite` platform requires at least Dragonite v1.5.0
- Params `routes`, `showcases`, `pokemon` and `gmf` were added separately to scanZone and scnaNext so you could set them as I did in default - scanZone is clearly used to scan mons while scanNext is used to scan routes, gmf and showcases details which may cause jump delays so it's proly worth to increase `userCooldownSeconds` in that case


server/src/services/api/scannerApi.js
```diff
@@ -124,6 +125,18 @@ async function scannerApi(
...
+      showcases: config.scanner.scanNext.showcases,
+      pokemon: config.scanner.scanNext.pokemon,
+      gmf: config.scanner.scanNext.gmf,
+    }
+    const scanZoneOptions = {
+      routes: config.scanner.scanZone.routes,
+      showcases: config.scanner.scanZone.showcases,
+      pokemon: config.scanner.scanZone.pokemon,
+      gmf: config.scanner.scanZone.gmf,
+    }
```
TurtIeSocks, line 139: Please move these lines up above the function declaration (where `backendConfig` is defined), since they're constant and don't need to be redeclared for every request. Also please use the same config syntax as the `backendConfig` definition uses. 

e.g.
```js
    const scanNextOptions = {
      routes: config.getSafe('scanner.scanNext.routes'),
      ...
    }
```

server/src/configs/default.json
```diff
@@ -595,8 +595,9 @@
   },
   "scanner": {
     "backendConfig": {
-      "platform": "rdm/mad/custom",
+      "platform": "rdm/mad/dragonite/custom",
       "apiEndpoint": "http://ip:port/api/",
+      "apiQueueEndpoint": "http://ip:port/api/",
```
TurtIeSocks, line 600: Is this necessary? You can't just add `queue` to the end like we've done in the past?

TurtIeSocks (changes requested):
Overall seems fine. I haven't familiarized myself with the V2 scout API but is having a separate `apiQueueEndpoint` config value necessary? Left a couple of comments to adjust as well. Thanks and good job.

lenisko (approved):
lgtm, thanks for changes

TurtIeSocks (approved):
Looks great now, thanks!

PR #966, opened 2024-02-23
refactor: move distance unit to client

### LINKED ISSUE #865: Get rid of miles
This option needs to die. Literally everything in the game uses (kilo)meters, notably routes which is user visible. RM also doesn't respect this option for routes.

_Originally posted by @Mygod in https://github.com/WatWowMap/ReactMap/issues/491#issuecomment-1767571170_

TurtIeSocks:
Moves the distance selection to the client for more consistency and there's not really any reason why this calculation should be forced to be the same for all users. 

## Config Change
`map.misc.distanceUnit` was previously `km/mi` and is now `kilometers/miles`, an inline migration was added to change it at startup with a warning log. 

Resolves #865

### DIFF
--- a/packages/locales/lib/human/en.json
+++ b/packages/locales/lib/human/en.json
@@ -734,5 +734,8 @@
   "gender_filters_all": "Gender Applies to All",
   "and_caption": "Linked Together",
   "or_caption": "Filters Independently",
-  "filter_help": "Filter Help"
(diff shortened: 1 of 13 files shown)


kamieniarz:
@TurtIeSocks `kilometers` and `miles` translations don't work

PR #967, opened 2024-02-26
feat: data management page

TurtIeSocks:
This adds a data management page that allows users to reset individual categories of what's saved in their local storage when things don't seem to be working instead of being forced to reset everything at once.

Screenshot 2024-02-26 at 3 10 03 PM


kamieniarz:
Filters could match perms, eg. if user doesn't have access to devices they shouldn't have them listed

TurtIeSocks:
It already does that.

PR #970, opened 2024-02-29
fix: no more search hard limits

### LINKED ISSUE #956: Missing results from search
Since recently, a lot of things seem missing when searching for them. For example, searching for a specific gym name could not locate that gym.

Mygod:
- fixes #956

i'm so nice

### DIFF
--- a/server/src/services/DbManager.js
+++ b/server/src/services/DbManager.js
@@ -436,7 +436,21 @@ class DbManager extends Logger {
     let deDuped = []
     let count = 0
     let distance = softLimit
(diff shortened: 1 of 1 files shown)


TurtIeSocks (changes requested):
Thanks for the PR! You're welcome to implement a more efficient algorithm if you'd like (or even a library), but since this PR currently undoes an intentionally implemented behavior for performance, it's a no from me.

Mygod:
Tested this as well. Since there is a timeout of 2 seconds, this should generally be fine?

TurtIeSocks (approved):

PR #975, opened 2024-03-02
Not fetch geo data if hidepii is true

mkllrtxn:
If logChannelHidePii is set to true do not fetch geo data from ip-api.com


TurtIeSocks (approved):
Thanks for the contribution!

PR #977, opened 2024-03-05
Add server cache to docker example

Fabio1988:
As it's not used for docker users if they don't add it manually.

Untested yet

### DIFF
--- a/docker-compose.example.yml
+++ b/docker-compose.example.yml
@@ -41,6 +41,7 @@ services:
       # ARRAY_VALUE_EXAMPLE: "[3, 4, 5]"
 
     volumes:
+      - .server/.cache:/home/node/server/.cache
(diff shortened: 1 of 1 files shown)


docker-compose.example.yml
```diff
@@ -41,6 +41,7 @@ services:
       # ARRAY_VALUE_EXAMPLE: "[3, 4, 5]"
 
     volumes:
+      - .server/.cache:/home/node/server/.cache
```
notheowner, line 44: `- ./server/.cache:/home/node/server/.cache`
Perhaps?

Fabio1988: Thanks, typo.

notheowner:
Looks good on my setup. No problems that I can see on my setup.

PR #990, opened 2024-03-25
Update GeoURI to drop a pin when opened with Google Maps

frett:
Wikipedia lists an unofficial extension using the `q` parameter to have Google Maps show a pin from a GeoURI. See: https://en.wikipedia.org/wiki/Geo_URI_scheme#Unofficial_extensions

I tested this on Android Google Maps and opening the link directly through the Tesla app.

I also tested the url pattern: `geo:0,0?q={x},{y}`, but opted to go with the committed pattern to account for external maps that may not recognize the `q` parameter


server/src/configs/default.json
```diff
@@ -907,7 +907,7 @@
     },
     {
       "name": "GeoURI",
-      "url": "geo:{x},{y}"
+      "url": "geo:{x},{y}?q={x},{y}"
```
frett, line 910: @TurtIeSocks I'm assuming you can use the replacement variables multiple times in a pattern

TurtIeSocks: Not currently but if you change https://github.com/WatWowMap/ReactMap/blob/b309e11ba18b41b91cdaae6d07af02a9d2051423/src/components/popups/Navigation.jsx#L18
to use `replaceAll` instead of `replace` (or a regex) then it shouldn't be a problem.

TurtIeSocks:
Thanks for the PR but can you change the target to `develop`? I'll sync them up in the very near future as I don't have much planned for `develop` currently.

frett:
give me a second to rebase and remove the other changes that are from master :)

TurtIeSocks (approved):
Looks great, thank you for the contribution!

PR #994, opened 2024-03-29
Change default minCp of Ultra

Astu04:
Skarmory Ultra Rank 1 CP is 2383

image
[Link to stadiumgaming.gg](https://www.stadiumgaming.gg/rank-checker?min_iv=0&league=great&att_iv=%2715%27&def_iv=%2715%27&hp_iv=%2715%27&pokemon=SKARMORY)

### DIFF
--- a/server/src/configs/default.json
+++ b/server/src/configs/default.json
@@ -92,7 +92,7 @@
       "minCp": {
         "little": 400,
         "great": 1400,
-        "ultra": 2400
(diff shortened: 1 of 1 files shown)


TurtIeSocks:
ReactMap is not supposed to account for hundos as PVP Perfect. They're just perfects. The `minCp` settings are also not used by Golbat 🤔 I think this one is fine to only set in your config.

PR #995, opened 2024-04-01
fix: locate translations / unit of measurement

TurtIeSocks:
It seems the translating API was changed when the version of Leaflet Locate Control was updated. This fixes the translations and also now respects the users choice of Miles (feet) or Kilometers (meters). Also fixes some React re-rendering issues and generally makes things more consistent.


Rodaviva29:
Works great, thanks!

PR #996, opened 2024-04-01
Add RoleLink info to Blocked page

XerockXMG:
This PR simply adds a button to link to the URL users have set in their config for `rolesLink` with the button text to be `rolesLinkName`.

Default color is `success` and also is set to open in a new tab.


TurtIeSocks (approved):
Works great, thanks for the contribution!

PR #1004, opened 2024-04-16
feat: added "basic" level gym badge

RradLaw:
This PR adds the gym badge level "basic" (which is the level before Bronze, but still different from "not obtained"). This better matches the options of gym badges in-game.

The changes in file `src/features/gym/gymMarker.js` allows the badge icon to appear with no border, which is how it is displayed in Pogo.

Previous gym badges/filters: All, None (badge_0), Bronze (badge_1), Silver (badge_2), Gold (badge_3)
Updated gym badges/filters: All, None (badge_0), Basic (badge_4), Bronze (badge_1), Silver (badge_2), Gold (badge_3)

image

image

image


src/features/gym/gymMarker.js
```diff
@@ -85,17 +85,21 @@ export function gymMarker({
...
-                  style="
-                    width: 48px;
-                    height: 48px;
-                    bottom: ${2 + gymMod.offsetY}px;
-                    left: ${gymMod.offsetX * 50}%;
-                    transform: translateX(-50%);
-                  "
-                />`
+                ${
+                  badge != '4'
```
TurtIeSocks, line 89: Whenever possible I try to do "positive" ternary checks to try and keep it consistent and clearer so let's change this to `badge == '4' ? '' : ...etc`

TurtIeSocks (changes requested):
Awesome, thank you for the contribution! You didn't have to add all of the locales, as they're generated when merged to main but you saved me some OpenAI credits, so thanks for that! @Petap0w, @ReuschelCGN / @Fabio1988, and @lenisko if you can please provide feedback on your respective human translations though that would be great!

I left one comment in the files changed, but it's nothing major, just code format nitpicking. You'll need to do something similar in the User Profile dialog as well as you currently end up with this:
Screenshot 2024-04-16 at 9 47 23 AM

Fabio1988:
I would suggest to use a different number, at some point we will have platin badges as well, what then? :)
Can we try to add it in between and add a migration on stored badges?

packages/locales/lib/generated/de.json
```diff
@@ -469,6 +469,7 @@
   "badge_1": "Bronze",
   "badge_2": "Silber",
   "badge_3": "Gold",
+  "badge_4": "Einfach",
```
Fabio1988, line 472: "Basis" @ReuschelCGN ? :)

RradLaw:
Thanks for the comments. I've fixed the profile page and updated the ternary.

Can you please point me to the right piece of code for the database migration?

RradLaw:
I've added the migration now and fixed the badges in the profile. Hopefully this should be all good now!

For the badge icons, I'm using `badge_${badge - 1}` to reference the gym icons so the icon repositories do not need to be updated.

image

packages/locales/lib/human/de.json
```diff
@@ -479,9 +479,10 @@
   "without_ar": "Ohne AR",
   "both": "Beide",
   "badge_0": "keiner",
-  "badge_1": "Bronze",
-  "badge_2": "Silber",
-  "badge_3": "Gold",
+  "badge_1": "Einfach",
```
Fabio1988, line 482:
```suggestion
  "badge_1": "Basis",
```

packages/locales/lib/human/pl.json
```diff
@@ -39,10 +39,10 @@
   "backups": "Kopie zapasowe",
   "backups_subtitle": "Umożliwia tworzenie kopii zapasowych i ładowanie ustawień na różne urządzenia",
   "badge_0": "Brak",
-  "badge_1": "Brąz",
-  "badge_2": "Srebro",
-  "badge_3": "Złoto",
-  "basic": "Podstawowy",
+  "badge_1": "Podstawowy",
```
lenisko, line 42: Please change it to `Podst.` with a dot at the end.
Otherwise this wil become too long.

lenisko (changes requested):

TurtIeSocks (approved):
Sorry for the delay, this is looking great now!

kamieniarz:
Something is wrong in profile tab with counters
image

RradLaw:
> Something is wrong in profile tab with counters image

What's the issue with it? It looks like the 4 badges in your image are "basic" level gym badges (not Bronze). Were they previously Bronze and the data migration did not work?

kamieniarz:
> > Something is wrong in profile tab with counters image
> 
> What's the issue with it? It looks like the 4 badges in your image are "basic" level gym badges (not Bronze). Were they previously Bronze and the data migration did not work?

I just added them and they counted as bronze. I also have way less gold and silver ones

PR #1006, opened 2024-04-18
Remove flex container behavior from AreaGroups

Cronick:
Remove container from AreaGroups, so in case you have Parent area(s) which doesn't fill out the entire Webhook popup width, then it align with the rest of them.

Before:
image

After:
image


TurtIeSocks (approved):
Good find and thanks for the contribution!

PR #1014, opened 2024-05-18
feat: add config to set scanner cooldown by role

SherlockSometimes:
Add discordRoleCooldown to scanNext and scanZone config to set cooldown per user role

example config (scanZone accepts the same format)
 "scanNext": {
      ...
      // Discord roles that will be allowed to use scanNext
      "discordRoles": ["Supporter","Scanner", "Mod"],
      // Default cooldown if no role specific cooldown
     "userCooldownSeconds": 60,
      //List of role/cooldown pair shortest matching time will apply
     "discordRoleCooldown":[["Supporter",30],["Mod",0]],
      ...
}

Expected outcome:
Supporter: 30 second cooldown
Mod: 0 second cooldown
Scanner: will get the default 60 second cooldown.
Supporter & Mod role: the lower 0 second cooldown would apply.


TurtIeSocks (changes requested):
Thanks for the contribution, I'm sure many are interested in this! I've been contemplating since you posted in the Discord for how we want to approach this in a scalable way that makes sense, but I keep ending up with solutions that add a lot of complications to this PR. So I'm going to try and keep this simple-ish for now and I may revise it in the future.

- Let's change the new config keys to just `rules`, as it will be used for Discord and Telegram. And I will use this more in the future as well...
- For the values themselves, we should switch to objects instead of a tuple, to be more consistent with the rest of the config / easier to read. Something like:
```json
{
  "rules": [
      { "role": "12415", "cooldown": 5 },
      { "role": "mod", "cooldown": 1 }
   ]
}
```
- In `server/src/services/config.js`, after replacing the alias roles with their ID counterparts, lets create a new map in each of the scanner modes. Something like:
```js
config.scanner.scanNext.rulesObj = Object.fromEntries(config.scanner.scanNext.rules.map((rule) => [rule.role, rule.cooldown]))
```
- Same for ScanZone
- In `server/src/services/DiscordClient.js`, remove the code you have now. Add:
```js
    // put this below the scanner declaration
    perms.scannerCooldowns = {}
    // Replace the code where the `scannerPerms` function is being called with
              scannerPerms(userRoles, 'discordRoles', trialActive).forEach(
                (x) => {
                  permSets.scanner.add(x)
                  perms.scannerCooldowns[x] = scanner[x].rules.reduce(
                    (acc, rule) => {
                      if (rule.cooldown < acc) {
                        return rule.cooldown
                      }
                      return acc
                    },
                    scanner[x].userCooldownSeconds,
                  )
                },
              )
```
- Add this above code to `server/src/services/TelegramClient.js` as well, with the appropriate adjustments for TG
- In `server/src/graphql/resolves.js`:
  - For the scannerConfig resolver, change these: 
  - `cooldown: perms?.scannerCooldowns?.[mode] || scanner.scanZone.userCooldownSeconds`
  - `cooldown: perms?.scannerCooldowns?.[mode] || scanner.scanNext.userCooldownSeconds`
  - You also need to adjust the `scanner` resolver, this one seems to have somehow escaped my attempt to have them in alphabetical order, so it's last in the `Query` object... Add something like:
```js
const cooldownSeconds = perms?.scannerCooldowns?.[category] || config.getSafe(`scanner.${category}.userCooldownSeconds`)
```
  - Then adjust the cooldown variable calculation to use this new value
- I would also like if you could adjust the `Permissions` type found in `packages/types/lib` so it has the new property. 

I know you mentioned on Discord that JS isn't your preferred language so let me know if you want me to jump in and help.  Thanks!

SherlockSometimes:
Didn't add the telegram part, I could add it but it would be untested as I don't use telegram, I can look at it tommorow if so, but figured id post the tested code and do that seperatly.
Wasn't sure if 'rules' needed to be added to 'Permissions' also as its part of the permSet object, so left it out for now.
Also added admin users get 0 cooldown as they skip the permission loop and would just get the default value.

packages/types/lib/server.d.ts
```diff
@@ -174,6 +174,7 @@ export interface Permissions {
   scanner: string[]
   areaRestrictions: string[]
   webhooks: string[]
+  scannerCooldowns: string[]
```
TurtIeSocks, line 177:
```suggestion
  scannerCooldowns: Record<string, number>
```

server/src/graphql/resolvers.js
```diff
@@ -384,15 +384,19 @@ const resolvers = {
...
               enabled: scanner[mode].enabled,
             }
           : {
               scannerType: scanner.backendConfig.platform,
               showScanCount: scanner.scanNext.showScanCount,
               showScanQueue: scanner.scanNext.showScanQueue,
-              cooldown: scanner.scanNext.userCooldownSeconds,
+              cooldown:
+                perms?.scannerCooldowns?.[mode] ||
+                scanner.scanZone.userCooldownSeconds,
```
TurtIeSocks, line 399:
```suggestion
                scanner.scanNext.userCooldownSeconds,
```

server/src/services/DiscordClient.js
```diff
@@ -129,17 +129,23 @@ class DiscordClient {
       webhooks: new Set(),
       scanner: new Set(),
       blockedGuildNames: new Set(),
+      rules: new Map(),
```
TurtIeSocks, line 132: I don't think we need this anywhere?

server/src/services/DiscordClient.js
```diff
@@ -129,17 +129,23 @@ class DiscordClient {
...
         Object.keys(this.perms).forEach((key) => (perms[key] = true))
         perms.admin = true
         config.getSafe('webhooks').forEach((x) => permSets.webhooks.add(x.name))
-        Object.keys(scanner).forEach(
-          (x) => scanner[x]?.enabled && permSets.scanner.add(x),
-        )
+        Object.keys(scanner).forEach((x) => {
+          if (scanner[x]?.enabled) {
+            permSets.scanner.add(x)
+            scanner[x].userCooldownSeconds = 0
```
TurtIeSocks, line 146:
```suggestion
            perms.scannerCooldowns[x] = 0
```
We want to assign this value to the local perms obj, not the config ref.

TurtIeSocks (changes requested):
Thanks for the adjustments, left some suggested changes on a few small things. 

As far as TG, if you just translate the code over to the TelegramClient then I will make an attempt to test it later this week. TG is quite a pain in the ass to test if you don't have already have it set up so I wouldn't ask anyone to go through that 😂

SherlockSometimes:
Switched to nullish ?? operator as 0 == false and would set default value when a 0 cooldown role was set.
Added telegram, it compiled and ran but no other testing was done with it.

kamieniarz:
Hey, I tried to use this PR by updating files manually but after all I could manage cooldown to disappear but in fact when I try to request a new scan within original cooldown timerange I have an error untill that 'virtual' (or actually hidden) original cooldown ends. Any chance this could be updated to latest version? Thanks

SherlockSometimes:
Did a merge with latest, don't have time to test it today, if you want you
can try it out, I'll try to get around to testing it tomorrow.

kamieniarz:
That's the same what I manually did in my code so the effect is the same as what I explained above

SherlockSometimes:
what does your config look like? seems to be working for me. a rule needs
to be added to scanZone/scanNext

    "scanZone": {
      "userCooldownSeconds": 30, // default cooldown
      "rules": [{ "role": "Supporter", "cooldown": 20 },
                   { "role": "SuperRole", "cooldown": 10 }],

SherlockSometimes:
just realized I never updated the original comment with the format change,
updated. Guess i need to work on adding telegram support and hopefully get
this merged in.

kamieniarz:
Here's examply
```
      "userCooldownSeconds": 20,
      "rules": [{ "role": "ADMIN", "cooldown": 1 }],
```
My map shows no cooldown but when I scan and try another scan (right after first one) I get an error

SherlockSometimes:
I see, I am also seeing the generic   "scan_error": "There has been an
error while processing the scan request...",
but I'm not seeing scan_error called in the code anywhere, no error is
printed to reactMap log, the chrome console, or dragonight.
So not even sure what the error is, I'll try to look around and see if I
can narrow down the error to find what is causing it.

PR #1048, opened 2024-09-04
fix: properly support gym defender temp evolutions

Mygod:
Gym defender will properly show reverted form after mega/primal evolution finishes.


TurtIeSocks (approved):
Thanks!

PR #1073, opened 2024-11-16
fix: increase wild level to 50

Mygod:


Fabio1988:
The 55 might have been a bug I guess, or it's intended who knows :D

Mygod:
I think it's better to allow them to show up on map when they get turned on.

Mygod:
image
image

Seems working lel.

PR #1094, opened 2025-03-12
feat: quest_shiny_probability

Mygod:
image

Support for https://github.com/UnownHash/Golbat/pull/267.

@jfberry @Fabio1988 I can't seem to find a quest with this field though?

### DIFF
--- a/packages/locales/lib/human/en.json
+++ b/packages/locales/lib/human/en.json
@@ -376,6 +376,7 @@
   "with_ar": "With AR",
   "both": "Both",
   "without_ar": "Without AR",
+  "shiny_probability": "Shiny: {{p}}",
   "exclude_quest_multi": "Exclude {{reward}}",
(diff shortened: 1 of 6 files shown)


Fabio1988:
They might add and remove that toggle from time to time.... It was present for gotour :)
Maybe they just need to send it during some certain events?

jfberry:
Maybe we should not include it when it is 0? If that is what it looks like when empty

Mygod:
It is indeed annoying. It looks like there isn't a way to distinguish missing `ShinyProbability` versus it being zero in the proto?

jfberry:
Or alternatively because they are paid for quests it may be ready to reveal odds under eu gambling laws

Mygod:
Yes but they probably shouldn't only reveal it in a place that can only be seen through data mine?

jfberry:
I’m thinking that this and other places (like eggs) which are related to purchases will start to comply with the law and show odds

PR #1097, opened 2025-03-14
fix: quest_shiny_probability pokestop popup and add german translation

ReuschelCGN:


src/features/pokestop/PokestopPopup.jsx
```diff
@@ -129,7 +129,7 @@ export function PokestopPopup({
                       }}
                     >
                       <QuestConditions {...quest} />
-                      {quest.quest_shiny_probability && (
+                      {!!quest.quest_shiny_probability && (
```
Mygod, line 132: 👀 What is this fixing? Was it not working before?

src/features/pokestop/PokestopPopup.jsx
```diff
@@ -538,7 +538,7 @@ const readableProbability = (x) => {
   const x_1 = Math.round(1 / x)
   const percent = Math.round(x * 100)
   return Math.abs(1 / x_1 - x) < Math.abs(percent * 0.01 - x)
-    ? `1/${x_1}`
+    ? `1:${x_1}`
```
Mygod, line 541: I'm fine with either but is there a reason to prefer this?

src/features/pokestop/PokestopPopup.jsx
Fabio1988: See comments in Discord, weird symbols :)

PR #1098, opened 2025-03-17
feat: add role array support

AppIe314:
Minor change to make it a little better to handle lots of Discord permissions in the local.json file.
Previously we would have to define roles like this:
```json
    "aliases": [
      {
        "role": "123",
        "name": "supporter-bronze"
      },
      {
        "role": "456",
        "name": "supporter-silver"
      },
      {
        "role": "789",
        "name": "supporter-gold"
      }
 ```
Now we can use an array to define a group of roles like this:
```json
    "aliases": [
     {
       "role": ["123", "456", "789"],
       "name": "supporter"
}
 ```

This makes permissions easier to read and handle, especially with many roles across different servers:
```json
"perms": {
      "iv": {
        "enabled": true,
        "trialPeriodEligible": false,
        "roles": ["supporter"]
      }
```

Tested on my own production instance.

### DIFF
--- a/packages/config/lib/mutations.js
+++ b/packages/config/lib/mutations.js
@@ -279,9 +279,9 @@ const applyMutations = (config) => {
   )
   /** @param {string | string[]} role */
   const replaceAliases = (role) =>
-    Array.isArray(role) 
-      ? role.flatMap((r) => aliasObj[r] ?? r) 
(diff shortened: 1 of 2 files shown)


Mygod:
fix checks plz

AppIe314:
yessir, just fixed the lint check

PR #1110, opened 2025-05-19
Drop MAD support

lenisko:
Since it's annoying to move around code with legacy MAD support, I went through and removed it. This should make RM easier to extend.

~~So far untested~~ Would be nice to have someone to test it and give feedback.


lenisko:
Briefly tested and seems to work just fine. Waiting for others.

Mygod:
This is the worst PR ever. 😡

Mygod:
> Well tested at this point... :)

I will never believe you. :rage:

server/src/models/Pokemon.js
```diff
@@ -317,26 +259,13 @@ class Pokemon extends Model {
       }
     }
     // second query for pvp
-    if (!mem && queryPvp && (!isMad || reactMapHandlesPvp)) {
+    if (!mem && queryPvp && reactMapHandlesPvp) {
```
Mygod, line 262: Bruh why you so bad at logic

server/src/models/Pokemon.js
```diff
@@ -317,26 +259,13 @@ class Pokemon extends Model {
...
-        ])
-        .andWhereBetween(isMad ? 'pokemon.longitude' : 'lon', [
-          args.minLon,
-          args.maxLon,
-        ])
-      if (isMad && listOfIds.length) {
+        .where('expire_timestamp', '>=', ts)
+        .andWhereBetween('lat', [args.minLat, args.maxLat])
+        .andWhereBetween('lon', [args.minLon, args.maxLon])
+      if (listOfIds.length) {
```
Mygod, line 268: wtf?????

Mygod:
PR reverted. Too many bugs. You might as well do it from scratch.

lenisko:
Yeah, thanks for revert.

What can I say, my army of testers is just bad :)

Mygod:
Maybe be careful and not write logical bugs to begin with, so that your army wouldn't need to test all the edge cases. :rage:

Issue #1101, opened 2025-04-07
Support for Gym defenders


Fabio1988:
see: https://github.com/UnownHash/Golbat/pull/281

Fabio1988:
> How often are gym defenders scanned? It would be confusing to users if it is only scanned like once a day.

Very frequently ;)
Gyms are also updated every 15 minutes at least, so you will see :)

Mygod:
Fixed in develop via #1112.

PR #1111, opened 2025-06-01
Gym Defenders

sanderDijkxhoorn:
Unfortunately I have even less time than anticipated, I created a PR with my current styling.

Current view displays:
- mon icon
- mon name with CP/deploy CP under it
- heart with lines displaying how many rounds necessary to take out mon

Implementing #1101 

image
image
image
image

TODO:
- Icon and/or button translations for show gym defenders
- Position of the show gym defenders button
- Fixed sizing so the text won't start more to the left for example with Gallade


lenisko:
I made required style changes:

p-1748910304-firefox
p-1748910287-firefox

Mygod:
I guess the heart with lines isn't really indicative for exactly how many turns if it is sitting between the divider? Maybe an additional text indicator like 2/3 or similar would be better?

Mygod:
Does the line curve up if it is above the line or something like that? Maybe we can only display the lines if it is below the threshold, so 0 lines if health is full, only the top line if it is half health, and both lines if it is almost empty.

sanderDijkxhoorn:
image

Conditional hearts & CP translation coming

PR #1112, opened 2025-06-03
Gym defenders implementation

sanderDijkxhoorn:
Implementing #1101 

image
image
image

Latest design:

image

I made a little oopsy with the last PR... here is a cleaner version of the PR, no GitHub workflow commits.


Mygod:
By the way the threshold numbers are definitely not one/two thirds. I think it's 73% or something similar.

src/features/gym/GymPopup.jsx
```diff
@@ -116,6 +144,219 @@ export function GymPopup({ hasRaid, hasHatched, raidIconUrl, ...gym }) {
...
+          }}
+        >
+          <Title backup={t('unknown_gym')}>{gym.name}</Title>
+        </Grid>
+      </Grid>
+      <Grid container direction="column" spacing={1}>
+        {defenders.map((def) => {
+          const fullCP = def.cp_when_deployed
+          const currentCP = def.cp_now
+          const percent = Math.max(0, Math.min(1, currentCP / fullCP))
```
Mygod: The best way to fix this is probably change this.

```diff
-const percent = Math.max(0, Math.min(1, currentCP / fullCP))
+const percent = Math.max(0, Math.min(1, currentCP / fullCP * 1.25 - .25))
```

lenisko:
I made another small change, including hearth thing @Mygod mentioned: https://github.com/sanderDijkxhoorn/ReactMapCC/pull/3

sanderDijkxhoorn:
I merged it, could someone try to run this PR and check if all functions correctly?

lenisko:
@sanderDijkxhoorn I'm running it at top of https://github.com/WatWowMap/ReactMap/pull/1113. Everything looks good related to defenders.

Mygod:
By the way, do we want to display some of these other fields? `deployed_time,battles_won,battles_lost,times_fed`

src/features/gym/GymPopup.jsx
```diff
@@ -116,6 +143,222 @@ export function GymPopup({ hasRaid, hasHatched, raidIconUrl, ...gym }) {
...
+          <Title backup={t('unknown_gym')}>{gym.name}</Title>
+        </Grid>
+      </Grid>
+      <Grid container direction="column" spacing={1}>
+        {defenders.map((def) => {
+          const fullCP = def.cp_when_deployed
+          const currentCP = def.cp_now
+          const percent = Math.max(
+            0,
+            Math.min(1, (currentCP / fullCP) * 1.25 - 0.25),
```
Mygod: Also it is better to use `def.motivation_now` instead of `currentCP / fullCP`. It gets 1/10000 precision rather than at most 1/4008 (for maxed Slaking).

lenisko:
> By the way, do we want to display some of these other fields? `deployed_time,battles_won,battles_lost,times_fed`

Waste of space... Unless we would replace `Pokemon name & CP line` lines with something like:

`:mdiMedal: {battles_won} :mdiEmoticonSad: {battles_lost} :mdiFoodApple: {times_fed}`
`{deployed_time}`

src/features/gym/GymPopup.jsx
Mygod: Maybe we can also try to predict the motivation value in the client? We have the formula for decay rate.

```js
const decayTime = 72 * 60 * 60 * Math.min(Math.max(Math.log10(3000 / fullCP), 1 / 9), 1)
const predictedMotivation = Math.max(0, def.motivation_now - Math.max(0, now - gym.updated) / decayTime)
const currentCP = Math.round(fullCP * (0.2 + 0.8 * predictedMotivation))
```

Source: https://www.reddit.com/r/TheSilphRoad/comments/at26qv/the_motivation_decay_algorithm_used_in_gyms_has/

Also do we want to update predicted CP every second? A maxed Slaking would decay by 1 CP roughly every 7.2 seconds.

jfberry: the json includes the back calculated deployed time at time of creation, and also the deployed ms - so you can see how old the json is if you need to (it will not be the same as gym.updated)

Mygod: Ah ok so the "Last Updated" time here in this PR is wrong as well.

jfberry: Could be, I haven't looked at the code.  For clarity, the gym updated column will show the last time the record was updated for any reason. Since the record is composed of data from a variety of protos, this can't be used to determine when the gym defenders were last updated.

However, knowing that the proto data for time deployed (elapsed=duration) wasn't useful without a baseline I added code to calculate based on when the json is written:
```golang
			DeployedMs: protoDefender.DeploymentTotals.DeploymentDurationMs,
			DeployedTime: now.
				Add(-1 * time.Millisecond * time.Duration(deploymentTotals.DeploymentDurationMs)).
				Unix(), // This will only be approximately correct
```

(the approximation here should only be a few milliseconds out).
If you actually want the time the gym defenders json was updated, then you can reverse this.  You also know how long it had been deployed at the time that the motivation was captured since the proto deployedms

Mygod:
image

Looks fucking good. UI done by vibe coding. @sanderDijkxhoorn @lenisko Merge?

sanderDijkxhoorn:
> image
> 
> Looks fucking good. UI done by vibe coding. @sanderDijkxhoorn @lenisko Merge?

Those icons look a little odd, replace it with berry and such?

Why does the top one start with 1d xxxx
The others have some hour minute seconds?

Mygod:
Minor complaints about UI will be only accepted in the form of commits or PRs. :)

lenisko:
@Mygod this looks like a :poop: We just literary been fighting for any vertical space, and you just added a bunch of useless stuff on the same page.

That's why I proposed additional data on toggle/click.

Mygod:
@lenisko I just care about functionality. If you want to make it look less 💩, PR welcome. :)

PR #1115, opened 2025-06-13
Fix: Search, old pokestops, small cleanup

lenisko:
- fixes search
- fixes hiding old pokestops (switch wasn't used)
- small MAD remainings cleanup

Thanks @ReuschelCGN !

### DIFF
--- a/server/src/models/Nest.js
+++ b/server/src/models/Nest.js
@@ -164,7 +164,8 @@ class Nest extends Model {
    * @param {ReturnType<typeof import("server/src/utils/getBbox").getBboxFromCenter>} bbox
    * @returns {Promise<FullNest[]>}
    */
-  static async search(perms, args, distance, bbox) {
+  // eslint-disable-next-line no-empty-pattern
+  static async search(perms, args, {}, distance, bbox) {
     const { search, locale, onlyAreas = [] } = args
     const pokemonIds = Object.keys(state.event.masterfile.pokemon).filter(
       (pkmn) =>
--- a/packages/types/lib/server.d.ts
+++ b/packages/types/lib/server.d.ts
@@ -34,7 +34,6 @@ export interface DbContext {
   hasRewardAmount: boolean
   hasPowerUp: boolean
   hasAltQuests: boolean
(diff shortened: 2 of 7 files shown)


server/src/models/Nest.js
```diff
@@ -164,7 +164,8 @@ class Nest extends Model {
    * @param {ReturnType<typeof import("server/src/utils/getBbox").getBboxFromCenter>} bbox
    * @returns {Promise<FullNest[]>}
    */
-  static async search(perms, args, distance, bbox) {
+  // eslint-disable-next-line no-empty-pattern
+  static async search(perms, args, {}, distance, bbox) {
```
jfberry, line 168: Why add in a dummy parameter? Why not just remove it from the calling sites?

ReuschelCGN: Dosn't removing it from calling sites brakes this search?
https://github.com/WatWowMap/ReactMap/commit/adf09fcedf373127070de9e13264e0dbd4d08be3#r160294066
in **server/src/models/Gym.js**
  `static async searchRaids(perms, args, { hasAlignment }, distance, bbox) {`

jfberry: that's the point to my question  - if the parameter is never used why is it even provided?

lenisko: I agree with Jabes. Once I will get close to PC I'll drop that.

jfberry: Of course there may be some complex something or other that requires a whole lot of functions to have the same signature, all called by search - but I thought it was worth asking the question because it's a very odd thing to do :-)

Mygod: Why no fix? :(

Mygod: I guess this was used here in `DbManager`:

```js
        this.models[model].map(async ({ SubModel, ...source }) =>
          SubModel[method](
            perms,
            args,
            source,
            DbManager.getDistance(args),
            bbox,
          ),
        ),
```

Looks like this has always been here since 2022?

PR #1139, opened 2025-09-27
Add access details to profile page

kamieniarz:
Added support for custom database that stores user data like access type and access expiration date (unix)

image

Made with AI so might not be ideal. Works for me tho

Disclaimer: I have no idea what I'm doing


Petap0w:
FYI this already exists with using `database` > `settings` > `extraUserFields` which makes use of the `data` field of the `reactmapusers` table.
Typically:
```
        {
          "name": "Status",
          "fr": "Statut",
          "database": "status",
          "disabled": true
        },
        {
          "name": "End Date",
          "fr": "Date de fin",
          "database": "premium_expiry",
          "disabled": true
        }
```
Which stored in database as `data` = `{"status":"Premium","premium_expiry_timestamp":1759435838,"premium_expiry":"02/10/2025"}`

Would give you:
Capture d’écran 2025-09-27 à 22 11 19

Then your role / access service just needs to use this data field and update it for each user.

kamieniarz:
🤦 to me lol 😅 Thanks!

PR #1144, opened 2025-09-29
Faster extraUserFields

kamieniarz:
These fields in profile (configured in config: database.settings.extraUserFields) refresh on re-login only. I made them update on each profile window open
image

Disclaimer: PR created with AI. I have no idea what I'm doing

### DIFF
--- a/server/src/routes/rootRouter.js
+++ b/server/src/routes/rootRouter.js
@@ -180,6 +180,9 @@ rootRouter.get('/api/settings', async (req, res, next) => {
               req.session.save()
             }
           }
+          if (user.data !== undefined) {
+            req.user.data = user.data
+          }
         }
       } catch (e) {
         log.warn(TAGS.session, 'Issue finding user, User ID:', req?.user?.id, e)
@@ -214,6 +217,12 @@ rootRouter.get('/api/settings', async (req, res, next) => {
       }
     }
 
+    res.set({
+      'Cache-Control': 'no-store, no-cache, must-revalidate, proxy-revalidate',
+      Pragma: 'no-cache',
+      Expires: '0',
+      'Surrogate-Control': 'no-store',
+    })
     res.status(200).json(settings)
   } catch (error) {
     res.status(500).json({ error: error.message, status: 500 })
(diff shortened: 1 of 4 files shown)


Mygod:
Umm I am suggesting that something else (non-reactmap) might be the issue here. Have you tried what I suggested?

Mygod:
Please check if #1145 fixes the issue for you.

P.S. You can call it a programmer's intuition but it appears to me that your patch works around the problem rather than addressing the root cause. Here's what my codex found:

- Blocking – src/features/profile/ExtraFields.jsx:57: the mutation still submits the pre-change value captured from the hook, so the database never sees the freshly typed target.value. That’s why the user only ever sees the “old” data again after a reload/login. The change set does nothing to address this root cause—please switch the payload to value: target.value (and ideally wait for the mutation result before mutating local state) so the server
actually stores the update.
- Minor – src/features/profile/index.jsx:37: once the mutation is fixed the extra getSettings fetch on every profile open becomes redundant churn. It only rewrites auth.data, leaves the rest of the settings stale (e.g. new fields won’t appear), and adds latency; consider removing it instead of shipping unnecessary network work.
- Neutral – server/src/routes/rootRouter.js:214: adding the Cache-Control header is harmless and may help with an overly aggressive CDN, but it isn’t what was breaking the feature—the stale payload bug above is. Keep the header if you want stricter cache semantics, but don’t expect it to compensate for the client bug.

Mygod:
Does your patch work if you revert the changes in `server/src/routes/rootRouter.js`?

kamieniarz:
No

server/src/routes/rootRouter.js
```diff
@@ -214,6 +217,12 @@ rootRouter.get('/api/settings', async (req, res, next) => {
       }
     }
 
+    res.set({
+      'Cache-Control': 'no-store, no-cache, must-revalidate, proxy-revalidate',
+      Pragma: 'no-cache',
+      Expires: '0',
+      'Surrogate-Control': 'no-store',
+    })
```
Mygod: What about this? I still don't think this is necessary.

kamieniarz: Works without that part

PR #1161, opened 2025-11-25
Fix logic bug in loading non-clustered elements

ComplementaryPogo:
If over the limit, prior behavior was to give up and not render **anything**, even if that component was set to ignore clustering.


Mygod:
@codex review

PR #1180, opened 2025-12-15
feat: hidden persistence

lenisko:
Persistent "Hide" feature. 

Hides everything for 1h can be restored via Settings > Clean Hidden.

Tested, working.

### DIFF
--- a/packages/locales/lib/human/en.json
+++ b/packages/locales/lib/human/en.json
@@ -138,6 +138,8 @@
   "exclude_lure": "Exclude Lure",
   "timer": "Timer",
   "hide": "Hide",
+  "hidden_for_hour": "Hidden for an hour",
+  "clean_hidden": "Clean Hidden",
   "tier": "Tier",
   "slots": "Slots",
   "mega": "Mega",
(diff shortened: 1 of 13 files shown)


Mygod (changes requested):
If you intend “hide for hour” to apply only to Pokémon, we should scope it to Pokémon; if you intend it to work for all hidden entities, we should add an automatic refresh/expiry so the in-memory Set drops expired IDs.

`To continue this session, run codex resume 019b2474-8894-73f2-a399-a4a5be980d90`

Mygod (changes requested):
That commit only partially addresses it.

- It does clean expired entries once after load and updates the in-memory Set (src/pages/map/components/Container.jsx:42 + src/utils/hiddenEntities.js:77), so stale hides from a previous session can drop ~15s after you open the map.
- It still does not make hides expire during a long-running session: after an hour passes, nothing re-runs cleanup, so useMemory().hideList stays stale until reload.
- It’s actually worse than before on “another hide refreshes it”: addHiddenEntity no longer calls cleanOldEntries, so hiding something later won’t drop expired IDs either (src/utils/hiddenEntities.js:54).

If “hidden for an hour” is still the intended behavior, you’ll need either cleanup-on-add back, or a one-shot setTimeout scheduled for the next expiry (not a loop) to refresh hideList when entries age out.

lenisko:
> As a human, I personally think hiding for exactly one hour is too arbitrary and the use case is too niche.

The original idea was for Pokemons, since map tends to refresh in background. Guess I can just bring back those and drop the rest

PR #1195, opened 2026-03-19
feat: support parent-based areaRestrictions in config for areas.json

unseenmagik:
## 📝 PR Overview
- feature: non breaking, additional config option.

For those that still use an areas.json file (rather than Koji), you can add a new parent parameter in the features.collection which can be used to filter access to areas on ReactMap.

- Added a new optional `parent: string[]` field to `authentication.areaRestrictions` so you can grant access by parent area name (and automatically include all child areas).
- Updated areaPerms.js to:
  - accept `parent` in the restriction rule,
  - build a parent→child key map from loaded geojson,
  - and grant access for both the parent area and its children.
- Updated GraphQL area filtering (`scanAreas` / `scanAreasMenu`) to allow permissions to match by `feature.properties.key` (in addition to `name`/`parent`) so the filtering behavior stays consistent with the new parent-based key logic.
- Updated type definitions (config.d.ts) to reflect the new optional `parent` array.
- Updated the example config (local.example.json) with `parent: []` entries for documentation and schema clarity.


unseenmagik:
Thanks for the updates @Mygod i will test these out.

Mygod:
bruh why so many bugs 😠

unseenmagik:
> bruh why so many bugs 😠

I never thought it would have taken this many fix passes 😳

Mygod:
I think I may have let my agent be too free. Steering it. (Rip my tokens.)

Btw just in case you can't tell yet, I'm using your PR as a testing playground for my new skill. ;)

Mygod:
Ok done. Please test thoroughly. Do not make any mistakes.

Mygod:
My agent has decided that this PR cannot be merged unless a major refactor is done first.

> I would stop trying to salvage this as a mergeable feature PR. Either:
> 
> - drop parent restrictions entirely for now, or
> - do a real redesign first with one canonical area identifier model and explicit validation for ambiguous configs

Sorry to disappoint you but the codebase at this point probably can't handle more complexity.

Mygod:
Also rip to wasted tokens.

PR #1201, opened 2026-04-01
feat: add IANA timezone support to trialPeriod config

lenisko:
When a timezone (e.g. "Europe/Warsaw") is set in trialPeriod, start/end times are interpreted as local time in that timezone. Interval advancement preserves local time across DST transitions by decomposing UTC dates into local components via Intl.DateTimeFormat and re-resolving the offset each cycle, so e.g. 17:00 Warsaw stays 17:00 Warsaw regardless of CET/CEST shifts.

Timer uses chained setTimeout instead of fixed setInterval when timezone is set, so each delay adapts to the actual wall-clock gap (167h or 169h around DST boundaries for a 168h interval).

Without timezone set, behavior is unchanged (server-local time, fixed ms intervals).

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -739,6 +739,7 @@
             "minute": 0
           },
           "intervalHours": 0,
+          "timezone": "",
           "roles": []
         }
       },
@@ -767,6 +768,8 @@
             "hour": 14,
             "minute": 0
           },
+          "intervalHours": 0,
+          "timezone": "",
           "roles": []
         }
       },
--- a/packages/types/lib/config.d.ts
+++ b/packages/types/lib/config.d.ts
@@ -66,6 +66,8 @@ export type Config<Client extends boolean = false> = DeepMerge<
         trialPeriod: {
           start: TrialPeriodDate
           end: TrialPeriodDate
+          intervalHours: number
+          timezone: string
           roles: string[]
         }
         allowedGuilds: string[]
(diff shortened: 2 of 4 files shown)


Mygod:
This PR cannot be merged in its current form.

The main issue is cost versus value. It adds a substantial amount of custom timezone/DST scheduling logic across Timer and Trial just to support an optional trialPeriod.timezone convenience field. That is long-term maintenance burden in one of the easiest areas to get subtly wrong.

More importantly, the implementation is still not robust. The recurrence logic used when fast-forwarding old trial windows is not the same as the logic used by the runtime timers, which means DST-crossing windows can drift by an hour. If we are going to take on timezone-aware scheduling, it needs to be correct and much more contained than this.

At this point, I do not think the feature justifies the additional complexity. It is better to keep trial periods in server time / UTC than to merge a bespoke calendar system that we will have to maintain indefinitely. If IANA timezone support is ever required, it should come back as a smaller design with one dedicated scheduling helper and explicit DST-focused tests.

lenisko:
> More importantly, the implementation is still not robust. The recurrence logic used when fast-forwarding old trial windows is not the same as the logic used by the runtime timers, which means DST-crossing windows can drift by an hour. If we are going to take on timezone-aware scheduling, it needs to be correct and much more contained than this.

Check again.

Current version is flawed as hell due to time shifts and requires changing config twice time per year.

Your (AI, not yours) costs are worth getting it working like it should.

PR #1204, opened 2026-04-03
fix(webhooks): normalize Poracle Pokemon payloads

Mygod:
## Summary

This fixes the ReactMap Poracle Pokemon alert flow so create and edit requests send the correct payload shape without leaking UI-only state.

### What changed

- split local Pokemon UI state from the payload sent to Poracle
- use a normalized create payload for add-new saves
- use a normalized update payload for tracked alert edits
- preserve `uid` on updates so existing Poracle rules are updated instead of duplicated
- preserve `ping` and other editable Pokemon fields across create/edit flows
- keep `allForms` handling stable when multiple forms of the same species are involved
- avoid reusing stale temp state in a way that can turn creates into updates

### Result

- new Pokemon alerts no longer accidentally overwrite existing tracked rules
- edited Pokemon alerts keep targeting the correct Poracle entry
- PoracleWeb and the map stay aligned on the saved Pokemon rule state
- default/full-range values are normalized consistently instead of leaking as noisy UI state

## Testing

- `./node_modules/.bin/eslint src/features/webhooks/services/Poracle.js src/features/webhooks/Manage.jsx src/features/webhooks/tiles/TrackedTile.jsx`
- `yarn build`

`yarn build` completed successfully with the existing offline locale-index fallback warning.

## Notes

There is still a preexisting no-IV summary mismatch on `develop`: the advanced dialog preview can hide some no-IV range constraints even though they may still persist. That behavior was identified during review but is not introduced
by this branch.

### DIFF
--- a/src/features/webhooks/Manage.jsx
+++ b/src/features/webhooks/Manage.jsx
@@ -91,17 +91,16 @@ export function Manage() {
 
   React.useEffect(() => {
     if (!addNew.open && addNew.save && category !== 'human') {
-      const { tempFilters } = useWebhookStore.getState()
-      const values = Poracle.toLocalState(
-        category,
-        Object.values(tempFilters || {}).filter((x) => x && x.enabled),
-        useWebhookStore.getState().context.ui[category].defaults,
+      const {
+        tempFilters,
+        context: { ui },
+      } = useWebhookStore.getState()
+      const { defaults } = ui[category]
+      const enabledFilters = Object.values(tempFilters || {}).filter(
+        (x) => x && x.enabled,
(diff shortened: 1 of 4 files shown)


jfberry:
Not sure if you are going to take this forward, but PoracleNG has v2 tracking routes attempting to remove ambiguity, inconsistent type coercion, and magic numbers from the api. If you are fixing poracle routes it is worth considering whether it is time to adopt the v2 api (documentation described in through openapi yaml, as a bonus)

Mygod:
I no longer remember what this is.

PR #1228, opened 2026-07-16
feat: fort map-data consumer — gyms, stations & pokestops - with DNF filtering

jfberry:
## Deploying this (end users)

Two small config changes — one on Golbat, one on ReactMap. Once both are set, fort map-data (**gyms, pokestops, stations**) is served from Golbat's in-memory fort API instead of SQL, with **automatic SQL fallback** if Golbat is ever unreachable — so it's safe to enable on a dual (endpoint + DB) source.

**1 · Golbat** — run Golbat with [UnownHash/Golbat#385](https://github.com/UnownHash/Golbat/pull/385) (it already includes the reward type-20 quest decode from #382) and turn on the in-memory fort index. `fort_in_memory` is a **top-level** option and defaults to `false`:

```toml
# golbat config.toml  (top level, above the [sections])
fort_in_memory = true
```

Make sure your Golbat API `secret` is set. Sanity-check the endpoint is live:

```bash
curl -H "X-Golbat-Secret: <secret>" http://<golbat-host>:<port>/api/fort/available
# 200 + { "pokestops": {...}, "gyms": {...}, "stations": {...} }
```

**2 · ReactMap** — add `endpoint` + `secret` to the scanner **DB source** whose `useFor` includes your fort types (`gym`/`pokestop`/`station`). This is the same Golbat endpoint most instances already point their **pokémon** source at; the DB source simply becomes **dual** (endpoint + DB):

```jsonc
// config/local.json → database.schemas[]   (your existing scanner DB source)
{
  "host": "127.0.0.1", "port": 3306,
  "username": "...", "password": "...", "database": "golbat_db",
  "endpoint": "http://<golbat-host>:<port>",   // ← add
  "secret": "<golbat api secret>",             // ← add (must match Golbat's)
  // "httpAuth": { "username": "...", "password": "..." },  // ← optional, if Golbat is behind Basic auth
  "useFor": ["gym", "pokestop", "station", "spawnpoint", "weather", "..."]
}
```

Restart ReactMap — this is a **server-only** change (no client rebuild). Gyms/pokestops/stations now fetch from Golbat; markers, popups, filters and deep-links keep working; and if Golbat is down or `fort_in_memory` is off, each fort type transparently falls back to SQL on the same DB.

> Prefer to scope it tighter? Split a `useFor: ["gym","pokestop","station"]`-only copy of the DB source with the endpoint added and drop those types from the original — but it isn't necessary.

**Confirm it's live (logs):** each ~15-min availability refresh logs a line like `[GYM] loaded available from http://<golbat>:<port>/api/fort/available — …`, and map pans stop issuing fort SQL on endpoint-backed sources. A Golbat outage instead logs a fallback warn and reverts to SQL.

---

## What this does

Routes fort map data — **gyms, pokestops, and stations** (`getAll`, `getOne`, `getAvailable`) — through Golbat's in-memory fort API ([UnownHash/Golbat#385](https://github.com/UnownHash/Golbat/pull/385)) when a scanner source has an `endpoint`, with transparent SQL fallback. Map pans stop issuing fort SQL entirely on endpoint-backed sources, and Golbat's DNF filtering returns only the forts that will render.

### Architecture

- **Endpoint branch per model** (`mem` set on the source): `POST /api/<type>/scan` → per-record pure mapper → the *existing, unchanged* `secondaryFilter`. On any failure (503 / network / bad shape) the model logs and falls through to SQL — a dual source degrades gracefully.
- **DNF is a superset narrow; `secondaryFilter` guarantees exactness.** Pure backends (`server/src/filters/fort/*`) translate the map filters into Golbat DNF clauses (OR across clauses, AND within). The one hard invariant is *never under-return*; anything inexpressible (quest title/target conditions, gender, gym badges, time-window cutoffs) poisons to match-all or stays a client-side residual.
- **Exact key semantics** in the translation (form-exact pokemon pairs, per-reward-type clauses, amount-exact mega/stardust/xp, reward type-20 treated as mega, grunt-class exclusion subtraction, station liveness) so the DNF result matches what renders.
- **Access-control parity**: the endpoint paths reproduce the SQL path's gates — strict area-restriction denial, freshness, `secondaryFilter`'s per-perm field gating — and `getOne` returns only `{lat, lon}` like SQL.
- **Observability**: each fetch logs the narrowing (`DNF(n): X matched -> Y after secondaryFilter (-residual) | <clause shapes>`), so any translation gap is visible in production logs.
- **Deep-link parity**: an off-viewport `onlyManualId` fort is fetched via the by-id endpoint and joins the candidate set, mirroring SQL's `(bbox) OR id = ?`.

### Source shapes

A scanner source can be **pure-DB** (unchanged), **dual** (endpoint + DB — fort reads use Golbat, everything else and the fallback use the bound DB), or **pure-endpoint** (endpoint only, no DB — degrades gracefully where a query would need a DB). Enabling the endpoint on an existing DB source (the common case above) makes it dual.

### Verified live (dual source, endpoint vs SQL)

| Path | Result |
|---|---|
| Pokestop quests | 2373 forts scanned → **8 returned → 8 rendered** (−0 residual) |
| Pokestop invasions (+ exclude grunts) | 2806 scanned → **107 → 107** (−0) |
| Pokestop lures / showcases | validated |
| Stations (incl. `station_active`, gmax) | validated; inactive-viewing unaffected (match-all preserved) |
| Gyms (raids/teams/badges) | validated, residual 0 |

### Requirements

- Golbat with [UnownHash/Golbat#385](https://github.com/UnownHash/Golbat/pull/385) (`fort_in_memory = true`), **deployed before** this — Golbat's scan decoder ignores unknown request fields, so an out-of-date Golbat degrades to match-all/SQL rather than erroring.
- Per scanner source: `endpoint` + `secret` (and optional `httpAuth`) as for pokémon; DB config unchanged (used for fallback + un-migrated queries).

No test framework exists in this repo; verification was throwaway node goldens per pure module (mappers, DNF backends), `eslint`/`prettier`, a full correctness/security/concurrency review of the branch, and the live parity runs above.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jfberry:
## Validating in a test environment

**Scope:** this branch routes only **gyms** through Golbat (`getAll`/`getOne`/`getAvailable`). Pokestop markers/popups and stations still use SQL. Live validation needs Golbat #385 deployed with `fort_in_memory = true`.

**1. Golbat prereqs** — deploy #385, `fort_in_memory = true`; sanity-check (`S`=secret, `G`=golbat url):
```bash
curl -H "X-Golbat-Secret: $S" "$G/api/gym/available"
curl -H "X-Golbat-Secret: $S" -H 'Content-Type: application/json' -XPOST "$G/api/gym/scan" \
  -d '{"min":{"latitude":LAT1,"longitude":LON1},"max":{"latitude":LAT2,"longitude":LON2},"limit":1000,"filters":[]}'
curl -H "X-Golbat-Secret: $S" "$G/api/gym/id/<gym-id>"
```

**2. ReactMap** — `git checkout feat/fort-consumer && yarn install`, restart (server-only changes — no client rebuild).

**3. Config** — add `endpoint` + `secret` to the scanner **DB** source whose `useFor` includes `gym` (makes it a *dual* source). If you already did this for the Phase-1 pokestop test, gyms are already routed.

**4. Confirm the endpoint path (logs):**
- ReactMap, each ~15-min refresh: `[GYMS] [GYM] loaded available from Golbat endpoint …/api/gym/available — …`
- `getAll`/`getOne` are silent on success; a `[GYM] /api/gym/scan … falling back to SQL` warn = the endpoint failed → SQL.
- Golbat: `available-gyms built in …` each time getAvailable is hit.

**5. Functional** — gym + raid markers render; a popup shows team/slots/defenders and raid boss + moves + CP; the filter drawer populates teams, tiers (`r`), eggs (`e`), raid bosses.

**6. Golden parity (endpoint vs SQL):** `curl "$R/api/v1/available/gyms?current=1"` with `fort_in_memory` **on** (endpoint) vs **off** (SQL fallback); `diff` the two → expect match (bar time-sensitive active-raid keys).

**7. Fallback** — stop Golbat or set `fort_in_memory=false` → gyms should **still render** via SQL + the fallback warn appears (no hard endpoint dependency; a timeout falls back too).

**8. Area restrictions** — as an area-restricted user, confirm gyms appear **only within permitted areas** (exercises the new client-side `filterRTree`) — the one behavior that isn't a straight SQL mirror, so the most important to verify.

jfberry:
This is now in my production and I have verified the DNFs through manual testing

jfberry:
Both Codex findings addressed in f161a967:

- **Pokestop endpoint-source gates** (`Pokestop.js`): the endpoint `getAll` path now applies the `hideOldPokestops` freshness gate (and the `onlyLevels`/`power_up_level` gate — vestigial since power-ups are out of the game) on the mapped rows, matching what the SQL path and the gym endpoint path already do, so endpoint sources no longer render stale/wrong-level stops.
- **`Station.getDynamaxMons`**: now endpoint-aware — reads `stationed_pokemon` from the whole-record `/api/station/id/{id}` response when the source has an `endpoint` (fixing the `this.query()` throw on pure-endpoint station sources), falling back to SQL for dual/SQL sources.

Mygod:
The endpoint-backed fort paths contain an access-control bypass and several contract, filtering, and fallback regressions. These can expose restricted station data or omit valid stations, quests, and invasions despite successful lint and build checks.

Full review comments:

- [P1] Enforce strict-area denial before endpoint station scans — src/models/Station.js:785-785
  When `authentication.strictAreaRestrictions` is enabled, restrictions are configured, and a user has no assigned areas, this endpoint branch returns before the SQL path's `getAreaSql` guard. `filterRTree` returns true when both lists are empty, so the user receives every station in the viewport instead of `[]`; apply the strict-area guard before accepting endpoint rows.

- [P1] Stop sending the unsupported station_active filter — server/src/filters/fort/station.js:37-37
  For every station request except inactive mode, this builder emits `station_active`, but the Golbat `ApiFortDnfFilter` contract targeted here has no such field—the companion work adds `stationed_gmax`, not `station_active`. Huma rejects unknown request properties, so endpoint-only station sources return no markers and dual sources always fall back to SQL; add the Golbat field/evaluator or stop sending it.

- [P1] Make advertised type-20 mega rewards retrievable — server/src/models/pokestopAvailableMapper.js:100-104
  When `/api/fort/available` reports a type-20 TEMP_EVO_BRANCH_RESOURCE with a Pokémon, this advertises an `m...` filter, but `buildPokestopDnfFilters` admits only reward type 12 and `parseRdmRewards` leaves non-empty type-20 rewards as `u20`. Selecting the advertised mega key therefore drops the row before or during secondary filtering, including the current `m150-150` fallback; normalize and accept type 20 consistently in both scan stages.

- [P1] Treat endpoint incidents as confirmed-capable — server/src/models/Pokestop.js:900-900
  For a pure endpoint pokestop source, `DbManager` never populates `hasConfirmed`, even though scan rows carry `confirmed` and all lineup slots. Passing `undefined` here makes `onlyConfirmed` ineffective for selected invasion filters and makes confirmed `a...` filters use the grunt's possible encounter pool instead of the observed slots; the endpoint context and `getFilterContext` should explicitly model confirmed support.

- [P1] Preserve every active incident during DNF narrowing — server/src/filters/fort/pokestop.js:277-278
  When a stop has multiple active incidents, Golbat's fort DNF index exposes only one flat `IncidentCharacter`/`IncidentDisplayType`, while `with_incidents:true` returns the full incident array. Narrowing on these fields drops a stop whenever the selected invasion or event is not the one held in the flat lookup, so secondary filtering never sees a valid match; poison multi-incident filters to match-all or index all active incidents.

- [P2] Keep historical rarity queries on dual DB sources — server/src/services/DbManager.js:287-287
  When an existing scanner DB schema—normally using one source for Pokémon and forts—is made dual by adding `endpoint`, this assignment makes its Pokémon source `mem` truthy. `historicalRarity()` then substitutes `[]` for that bound source and clears the historical rarity map on refresh despite the DB still being available; distinguish a pure endpoint from a dual source in that skip condition.

- [P2] Apply result limits after local fort filters — server/src/models/Gym.js:542-542
  When a DNF scan has more than `queryLimits.gyms` candidates—for example, a large viewport where most candidates fail the user's area polygon—Golbat stops traversal at this limit before `filterRTree`, freshness checks, and `secondaryFilter` run. Valid matches later in the traversal are lost, unlike SQL's WHERE-before-LIMIT behavior; the same pre-residual cap is sent by pokestop and station scans, so push those gates upstream or defer the final cap until after local filtering.

- [P2] Return null on endpoint-only getOne misses — server/src/models/Gym.js:854-854
  When an endpoint-only source returns 404 or no response for a stale ID, `fetchFortById` returns null and execution reaches this SQL query. That model has no bound knex, so it throws; `DbManager.getOne` uses `Promise.all`, causing the whole `gymsSingle` request to fail—the identical fallback exists in `Pokestop.getOne`—instead of returning no match. Return null when there is no bound DB fallback.

- [P2] Include confirmed rewards from every rocket slot — server/src/models/pokestopAvailableMapper.js:176-177
  When `map.misc.fallbackRocketPokemonFiltering` is false, this mapper is the source of confirmed `a...` availability keys, but it reads only slot 1. The SQL path adds configured rewards from slots 1, 2, and 3, so endpoint-backed sources omit filters for confirmed rewards in later slots; expose and map all three slots.

jfberry:
Thanks for the thorough review, @Mygod. Walking through all nine — seven are addressed (across this PR and the Golbat companion #385), and two look like they were reviewed against a Golbat without #385's changes. Details below.

### Fixed

- **#1 — strict-area denial before endpoint scans (P1).** Confirmed bug: the endpoint paths filter with `filterRTree`, which returns `true` for empty area inputs, so a no-area user under `strictAreaRestrictions` was not denied the way the SQL `getAreaSql` denies. Added `areaRestrictionsDenyAll` (mirrors `getAreaSql`'s strict-deny and empty-consolidated deny) and short-circuit each fort model's endpoint branch to `[]`. — `15d82178`
- **#3 — type-20 mega retrievable (P1).** Root cause was upstream: Golbat's vendored proto didn't decode reward type 20 at all (empty `info`), which is what the `m150-150` empty-info stopgap worked around. With Golbat #382 merged (decodes `TEMP_EVO_BRANCH_RESOURCE` → `pokemon_id`/`amount`), type 20 is now treated as mega everywhere — SQL match/advertise/search, `parseRdmRewards`, and the endpoint DNF (`quest_reward_type: [12, 20]`) — and the GoFest stopgap is retired. — ReactMap `cade9f68`, Golbat `c3acb26`
- **#4 — endpoint incidents confirmed-capable (P1).** A pure-endpoint pokestop source now reports `hasConfirmed: true` (the Golbat scan always returns confirmed lineup data; no DB schema check runs for a knex-less source). — `f51f662e`
- **#6 — historical rarity on dual sources (P2).** `historicalRarity()` now skips only pure-endpoint sources (no bound knex) rather than any source with `mem`, so a dual source keeps contributing. — `f51f662e`
- **#7 — fort scan result limit (P2).** On the Golbat side, the fort scan traversal cap read `Tuning.MaxPokemonResults`; added a dedicated `max_fort_results` (default 9000) so the fort API can be tuned independently of pokemon. The per-request `Limit` still `min()`s it. — Golbat `a3c82e5`
- **#8 — `getOne` on endpoint-only misses (P2).** `DbManager.getOne` now uses `Promise.allSettled`, so a pure-endpoint source whose by-id fetch misses (falling through to an unbound `this.query()`) no longer fails the whole single-fort lookup. — `f51f662e`
- **#9 — confirmed rewards from every rocket slot (P2).** Golbat's invasion availability exposed only slot 1; it now exposes slots 2/3 (`FortLookupIncident`/`invasionKey`/`ApiPokestopInvasionAvailable`), and the mapper advertises an `a` key per slot the event marks (`second`/`thirdReward`). Index cardinality is unchanged — a confirmed grunt's lineup is deterministic per character, so `(character, slot1)` already determines slots 2/3. — Golbat `e8c4505`, ReactMap `fe24b481`

### Not reproduced (appear to predate Golbat #385)

- **#2 — `station_active` unsupported filter.** `station_active` **is** part of the Golbat `ApiFortDnfFilter` contract in the companion #385: it's a declared field with an evaluator (`isFortDnfMatch`: `*StationActive != (StationEndTimestamp > now)`) and a dedicated test. Huma rejects only *unknown* properties, and this one is declared — so endpoint station sources are not rejected. #385 adds both `stationed_gmax` **and** `station_active`.
- **#5 — multi-incident DNF drops stops.** `FortLookup.Incidents` is a slice (`[]FortLookupIncident`), and `isFortDnfMatch` iterates **all** non-expired incidents with match-any semantics, not a single flat incident. ReactMap also emits `incident_character` and `incident_display_type` as separate clauses (never AND-ed within one), so a stop with the selected invasion in any active incident still matches. If you saw a drop against #385's matcher I'm happy to dig into a specific repro.

All changes build/lint clean; Golbat decoder tests pass (incl. new slots-2/3 coverage). #3, #9, and the type-20 path assume a deployed Golbat ≥ #385 + #382.

jfberry:
I've taken the change to support slot2/slot3 availability and querying (which I previously had not added because these fields are always empty). Just like AR searching, these fields are legacy fields at the moment - but slot2/3 data coming back is something we would hope for though unlikely, whereas AR is something which is likely gone forever. A focused change later to drop AR data capture & reporting will be a wider impact in golbat

Mygod:
Type-20 workaround has been removed in https://github.com/WatWowMap/ReactMap/commit/9acf867e68b6cb7d595fa45d81d4275e75592162. Please merge main and resolve conflicts. Hopefully it will make things easier.

jfberry:
Merged `develop` (through `9acf867`) into the branch — thanks, that did make it simpler.

My type-20 handling has **converged on develop's**: I reverted my SQL-side type-20 generalization (it was redundant with `9acf867` + the reward-definition refactor) and took develop's `TEMP_EVOLUTION_RESOURCE_REWARD_TYPES` / `QUEST_REWARD_FILTER_DEFINITIONS` approach wholesale. The only fort-consumer-specific pieces that remain are:

- **Endpoint DNF** (`filters/fort/pokestop.js`): mega clauses emit `quest_reward_type: [12, 20]` so Golbat returns type-20 (temp-evo branch) mega stops for the in-memory scan — develop doesn't have this file.
- **`parseRdmRewards` guard**: the merge resolution keeps `typeof quest.quest_rewards === 'string' ? JSON.parse(...) : quest.quest_rewards` (+ an `Array.isArray` guard). The endpoint path hands `parseRdmRewards` an already-parsed rewards array rather than a JSON string, so develop's `JSON.parse(quest.quest_rewards)[0]` alone would throw on endpoint sources. Flagging it as the integration seam if `parseRdmRewards` gets touched again.

No conflict markers, lint clean; the DNF, type-20 key, and rocket slots-2/3 mapper pass their unit checks. This assumes deployed Golbat ≥ #385 + #382 for the type-20 / slots-2/3 paths.

Mygod:
The endpoint Pokestop path currently returns no markers because it omits a required filter limit, and one supported pure-endpoint configuration can fail during startup. Additional endpoint capability, availability, manual-ID, performance, and source-maintainability regressions also remain.

Full review comments:

- [P1] Pass the result limit to endpoint filtering — server/src/models/Pokestop.js:939-943
  For every successful endpoint-backed Pokestop scan, this call omits `secondaryFilter`'s final `resultLimit` argument. Its loop condition is `filteredResults.length < resultLimit`; with `undefined` this is false immediately, so even non-empty endpoint responses produce no markers. Pass `queryLimits.pokestops` as the final argument, as the SQL call below does.

- [P1] Keep pure endpoint filter context off the SQL path — server/src/services/DbManager.js:285-285
  When `fallbackRocketPokemonFiltering` is disabled and Pokestop uses a pure endpoint source, setting `hasConfirmed` here makes `Pokestop.getFilterContext()` bypass its `!hasConfirmed` return and execute `this.query()` on an unbound model. `state.loadLocalContexts()` awaits that during startup, so initialization rejects; make the filter-context path recognize `mem` and return endpoint capability without SQL.

- [P2] Preserve the appended manual Pokestop at the limit — server/src/models/Pokestop.js:929-931
  When the scan already returns `queryLimits.pokestops` candidates and `onlyManualId` points outside the bounding box, the by-ID record is appended to the end and then this truncation immediately removes it. Deep links therefore fail in full viewports; retain the manual row when capping or apply the final limit after secondary filtering.

- [P2] Drop incomplete type-20 rewards from availability — server/src/models/pokestopAvailableMapper.js:108-108
  When Golbat reports a type-20 tuple without both a positive `pokemon_id` and `amount`, this emits `u20` or `m{id}-0`, but `secondaryFilter` treats type 20 as a dedicated mega reward and drops it when its computed mega key is empty. The drawer consequently advertises a filter that can never return a marker; match the SQL path by adding type 20 only when both fields are present.

- [P2] Mark dual endpoint Pokestops as confirmed-capable — server/src/services/DbManager.js:292-295
  When a dual source's database lacks the `confirmed` column, `schemaCheck` leaves `hasConfirmed` false and this overlay copies only endpoint credentials, even though subsequent `getAll` calls use Golbat rows carrying confirmation and lineup slots. Consequently `onlyConfirmed` is ignored and reward filtering falls back to possible grunt encounters instead of the actual lineup; override this capability whenever the endpoint is active.

- [P2] Do not send unsupported station_active clauses — server/src/filters/fort/station.js:37-37
  For endpoint-backed station scans, Golbat's `ApiFortDnfFilter` has no `station_active` field and its JSON decoder silently ignores unknown members. In `onlyAllStations` mode this therefore becomes an empty clause matching every expired station, so the intended cache and payload reduction does not occur; add the corresponding Golbat predicate or leave liveness as an explicit residual match-all.

- [P2] Replace literal NULs in the cache key — server/src/utils/fortAvailable.js:31-31
  The separators in this template literal are literal NUL bytes, causing Git to classify the entire JavaScript file as binary so normal source diffs and some text tooling cannot inspect it. Use escaped separators such as `\0` in the source to preserve the runtime key while keeping the file reviewable.

Also please change the merge target to `main` branch, also there has been some additional changes introduced in the main branch meanwhile.

jfberry:
Thanks — all seven addressed, plus retargeted to `main` and merged the latest `main`.

**P1**
- **Result limit not passed to endpoint filtering** — the endpoint `secondaryFilter` call now passes `queryLimits.pokestops` (its loop is `filteredResults.length < resultLimit`, so `undefined` returned zero markers). Removed the pre-truncation entirely, mirroring the SQL call.
- **Pure-endpoint filter context off the SQL path** — `getFilterContext` now recognizes `mem` and returns `{ hasConfirmedInvasions: true }` without touching `this.query()`, so startup no longer rejects on an unbound model when `fallbackRocketPokemonFiltering` is off.

**P2**
- **Manual Pokestop at the limit** — fixed by the same change: no pre-truncation, so the appended off-viewport by-id row survives into `secondaryFilter` (parity with SQL's `OR id = manualId` + `LIMIT`).
- **Incomplete type-20 in availability** — `questRewardKey`/the mapper now emit `m<id>-<amt>` only when both `pokemon_id` and `amount` are present, and never `u20` (type 20 has a dedicated filter). Matches what `secondaryFilter` keys, so no dead drawer entries.
- **Dual endpoint confirmed-capability** — the dual overlay now also sets `hasConfirmed: true`, so `onlyConfirmed` and confirmed rocket-reward filters use the Golbat lineup even when the bound DB lacks the `confirmed` column.
- **Literal NULs in the cache key** — switched to the `\0` escape in source; git no longer classifies `fortAvailable.js` as binary.

**`station_active`** — I'll push back gently here: it *is* in the Golbat companion (#385) — a declared `ApiFortDnfFilter` field with an evaluator (`*StationActive != (StationEndTimestamp > now)`) and a test, not an unknown member. You're right that Huma silently ignores genuinely-unknown fields (no `additionalProperties: false`), but that's the general DNF-superset contract — any field a given Golbat doesn't recognize is ignored and `secondaryFilter`/the station time-gate still finalizes correctness (just without the payload reduction). Against a Golbat that predates #385's station changes it degrades to residual match-all rather than breaking. The consumer already requires #385, so it resolves optimally there. Happy to add a hard residual fallback if you'd rather not depend on the field.

Base is now `main`; note the diff currently carries `develop`'s commits that aren't yet in `main` — those drop out once `develop` lands on `main`.

jfberry:
## Whole-PR review + fixes

Given how much moved recently (the `develop`/`main` merges, type-20 convergence, two review rounds), I ran a full review across three dimensions — correctness/SQL-parity, security/access-control, concurrency/integration — over the ~2,050-line feature diff, verifying each finding against the code. **8 issues** surfaced (2 P1, 3 P2, 3 P3); 7 are fixed in `09ac0769`, 1 is a documented accept.

### P1 (fixed)
- **`getOne` returned the full Golbat record** — SQL `getOne` projects to `{lat, lon}`; the endpoint returned the raw record, and since the client controls the GraphQL selection, a `gyms`-but-not-`raids` user could read raid/team/detail for any fort id, outside their areas. Now projected to `{lat, lon}` to match SQL.
- **Pokestop `onlyLevels` under-return** — the endpoint applied the power-up gate to every row, but SQL applies it only under `onlyAllPokestops` (the gym sibling has the `!onlyAllGyms ||` guard). A user with a non-`all` levels filter lost the whole quest/invasion/lure layer. Added the `!onlyAllPokestops ||` guard.

### P2 (fixed)
- **`search` / `submissionCells` crashed for pure-endpoint sources** — those fort methods have no endpoint branch, and DbManager batched them with `Promise.all`, so a pure-endpoint source rejected the whole batch. Switched to `Promise.allSettled` (mirrors the `getOne` fix), degrading that source to no results.
- **Credentials written to disk** — `fetchJson` dumped request `options` (incl. `X-Golbat-Secret` / Basic auth) to `logs/` on any failed request. Now redacts those headers before the debug log and the payload dump.
- **Availability TTL armed on empty refresh** — a failed (empty) endpoint refresh armed the 60s TTL and suppressed recovery. TTL is now only armed on a non-empty result.

### P3
- **Fixed**: `encodeURIComponent` on all fort by-id fetch URLs (no path traversal via `id`/`onlyManualId`); `historicalRarity` → `allSettled` so a dual DB lacking `pokemon_stats` doesn't blank every source's rarity map.
- **Accepted (not changed)**: `force` doesn't bypass the 30s combined-availability cache — that cache deliberately caches failures to avoid hammering an old/down Golbat, and the TTL fix already cuts the drawer-empty window to ≤30s; a force-bypass needs invasive threading for a minor, drawer-only gain.

### Verified solid (no action)
- All three DNF builders correctly mirror SQL as superset-narrows — layer-gating, exact quest keys (form:0/bare, amount-exactness, type-20 `[12,20]`), rocket→`incident_character` expansion, `onlyGymBadges`→`[]`, station `station_active`. No under-returns.
- All availability mappers reproduce the SQL `getAvailable` keys (incl. type-20 complete-only, rocket slots 1/2/3).
- Strict-area guard is complete on all three `getAll` endpoints and faithfully mirrors `getAreaSql`; manual-id preserves area filtering; cache key is credential-isolated.
- Golbat's DNF matches **both** AR and non-AR quest layers (`arMatch || noArMatch`) — no systematic layer under-return.

`node --check` + ESLint clean across all changed files. As before, worth a functional smoke test on a pokestop source (quests + confirmed invasions) once pulled, given there's no server test suite. Requires Golbat ≥ #385 + #382.

jfberry:
(running in my production, will also ask my other testers of this PR to update after a short period of no further review comments)

Mygod:
The endpoint paths can omit valid forts, break advertised dual-source fallback, and replace cached availability or rarity with failure-derived empty data. Lint and production build pass, but they do not cover these runtime contract and failure-mode regressions.

Full review comments:

- [P1] Apply the fort result cap after residual filters — server/src/models/Pokestop.js:886-886
  When an endpoint-backed query uses residual-only conditions such as area restrictions, freshness, advanced quest conditions, gym badges, or station activity, Golbat stops after the broader DNF reaches its result cap. Rows later rejected by `filterRTree` or `secondaryFilter` consume that cap, so valid forts appearing later in the scan are omitted; use pagination/over-fetching or exact server-side predicates before enforcing the ReactMap limit. The same pre-filter cap exists in `Gym.js` and `Station.js`.

- [P1] Keep multi-incident stops out of flat DNF filters — server/src/filters/fort/pokestop.js:281-282
  When a stop has multiple simultaneous incidents and the selected character or display type belongs to a non-indexed incident, these clauses compare against Golbat's single flat `FortLookup` incident tuple. The scan drops the stop before `with_incidents` can return the full `invasions[]` that `secondaryFilter` would match; leave these predicates residual or make Golbat test every active incident.

- [P1] Preserve cached availability when refresh fails — server/src/services/EventManager.js:172-177
  When a pure-endpoint `/api/fort/available` request fails, `Db.getAvailable` returns `[]`, but this branch runs only after the cached options have already been replaced and Pokestop quest conditions reset. The drawer is therefore emptied instead of retaining the last good result; conversely, a genuinely successful empty result is never timestamped and is re-scanned on every session. Propagate refresh success separately and only commit or arm the TTL for a successful refresh.

- [P1] Preserve DB capability flags for endpoint fallback — server/src/services/DbManager.js:296-300
  For a dual endpoint-plus-DB source whose incident table lacks `confirmed` or lineup columns, this also changes the context used by the SQL fallback. If `/api/pokestop/scan` or `/api/fort/available` is unavailable, fallback queries reference `confirmed` and `slot_1/2/3_*`, reject, and are discarded by `runScannerSources`, leaving no pokestops or availability. Keep endpoint capabilities separate from schema-derived fallback flags.

- [P2] Remove the unsupported station_active clause — server/src/filters/fort/station.js:37-37
  For endpoint-backed station queries outside inactive mode, this emits `station_active`, but Golbat's `ApiFortDnfFilter` has no such field; the planned producer extension only adds `stationed_gmax`. The current decoder ignores the unknown key, so cases such as `onlyAllStations` become match-all and expired or stale stations can exhaust the endpoint's hard result cap before the local time gate. Keep liveness residual or add it to the producer contract first.

- [P2] Preserve conditions for generic quest rewards — server/src/models/pokestopAvailableMapper.js:172-175
  When Golbat reports a reward type mapping to `u<type>`, this branch omits its title and target from `conditions`. The SQL path calls `process` for every generic quest row, so endpoint-only deployments lose the advanced quest-condition selector for these rewards and existing `adv` selections can be cleared by the client. Run generic keys through the same condition collector.

- [P2] Retain historical rarity after total query failure — server/src/services/DbManager.js:375-377
  If every eligible `pokemon_stats` query rejects, such as during a temporary outage of the sole dual-source DB, filtering rejected promises leaves `results` empty and `setRarity` clears the cached historical map. Previously `Promise.all` entered the catch path and retained the last good data. Track whether any eligible query succeeded and avoid replacing rarity state on total failure.

jfberry:
Thanks — four of these were real ReactMap-side regressions and are fixed in 4b278a31; the other three reference the Golbat companion contract and are addressed there rather than by broadening the queries.

### Fixed (4b278a31)

- **Generic quest conditions** — the availability mapper now runs every key, including generic `u<type>` fallbacks, through `process()`, so they carry their title/target conditions to match develop's SQL `genericQuests.forEach(process)`. Endpoint deployments no longer lose the advanced quest-condition selector.
- **Preserve cached availability on failed refresh** — `#refreshAvailable` now returns early on an empty result, so a failed (empty) `/api/fort/available` no longer replaces the last-good drawer + conditions or arms the TTL; it keeps the cache and retries on the next call.
- **Endpoint capability vs. SQL-fallback flags** — reverted the dual-source `hasConfirmed = true` overlay (it poisoned the SQL fallback into querying a `confirmed` column the DB may lack). The endpoint scan branch now marks itself confirmed-capable *locally*, leaving the schema-derived flag for the fallback.
- **Retain historical rarity on total failure** — if every eligible `pokemon_stats` query rejects, the last-good rarity map is retained instead of cleared (restoring the old `Promise.all` catch-path behaviour that the `allSettled` change had dropped).

### Addressed on the Golbat side / not broadened

These three are contract points against Golbat [#385](https://github.com/UnownHash/Golbat/pull/385) (the companion), and #385 already covers them — so the fix is to rely on the contract, not to widen what Golbat returns:

- **Result cap before residual filters** — handled by a dedicated, separately-tunable Golbat cap (`max_fort_results`, default 9000, independent of `max_pokemon_results`) rather than over-fetching/pagination, which would broaden every scan. The residual gates (area polygons, freshness, adv conditions, badges) can't be pushed server-side, so the deliberate choice is a high, operator-tunable ceiling.
- **Multi-incident stops** — [#385](https://github.com/UnownHash/Golbat/pull/385)'s `FortLookup.Incidents` is a slice (`[]FortLookupIncident`), not a single flat tuple, and `isFortDnfMatch` iterates **every** non-expired incident (`for _, inc := range fortLookup.Incidents` → match-any). A stop matches if any active incident satisfies the clause, and ReactMap emits `incident_character` / `incident_display_type` as separate clauses (never AND-ed), so no valid stop is dropped.
- **`station_active`** — it is a declared field in [#385](https://github.com/UnownHash/Golbat/pull/385)'s `ApiFortDnfFilter` (`decoder/api_fort.go`) with an evaluator (`*StationActive != (StationEndTimestamp > now)`) and a test, so it's part of the producer contract — not an unknown key. Leaving liveness residual would instead ship every expired station and defeat the payload reduction. ([#385](https://github.com/UnownHash/Golbat/pull/385) adds both `stationed_gmax` and `station_active`.)

All four fixes are lint/`node --check` clean and the mapper's generic-condition behaviour is unit-verified. As always this assumes Golbat ≥ [#385](https://github.com/UnownHash/Golbat/pull/385) + [#382](https://github.com/UnownHash/Golbat/pull/382).

Mygod:
The endpoint paths can silently omit valid forts, lose manual deep links, and produce inconsistent availability state. They also introduce source-failure masking and an endpoint-only GraphQL type violation.

Full review comments:

- [P1] Apply scan limits only after residual fort filters — server/src/models/Pokestop.js:886-886
  When the endpoint DNF returns at least `queryLimits.pokestops` candidates, Golbat stops at this limit before `filterRTree` and `secondaryFilter` run. Candidates outside the user's permitted polygon or failing quest-layer/title/target checks therefore consume the cap, and valid later candidates are never fetched; page or fetch the DNF superset until the post-filter limit is filled. The Gym and Station endpoint limits have the same ordering problem.

- [P2] Keep rocket-reward DNF as a true superset — server/src/filters/fort/pokestop.js:259-261
  When confirmed lineup data gets ahead of the invasion metadata—for example, a rotation moves a reward Pokémon from one grunt type to another—`secondaryFilter` accepts the new grunt using its actual slot, but this expansion emits only the old metadata-derived `incident_character` values, so Golbat discards the valid stop first. Since actual slots cannot be expressed by this DNF field, an `a...` filter must disable this narrowing or use an API predicate over the confirmed slots.

- [P2] Put the manual Pokéstop before capped candidates — server/src/models/Pokestop.js:908-908
  When the scan already returns 5,000 matching stops and they all pass, the fetched manual stop is appended as element 5,001 while `secondaryFilter` stops after collecting `resultLimit` entries and never examines it. This makes an off-viewport `onlyManualId` deep link disappear in dense views; reserve a slot or place the manual result before ordinary candidates.

- [P2] Preserve manager-owned metadata on empty refreshes — server/src/services/EventManager.js:139-139
  When all Pokéstop availability sources fail, `Db.getAvailable('Pokestop')` has already replaced `Db.questConditions` with `{}` before returning an empty array, so this early return preserves only the old filter-key list while deleting its advanced condition metadata. The analogous Pokemon path can also clear rarity first; make these manager side effects transactional or restore them when rejecting an empty refresh.

- [P2] Invalidate the availability TTL when the manager changes — server/src/services/EventManager.js:110-115
  When database or endpoint configuration is hot-reloaded within this TTL, `state.reload` replaces `state.db` and immediately calls `loadLocalContexts`, but this category-only timestamp returns without querying the replacement manager. Availability and manager-owned metadata consequently remain tied to the old source; key the cache by manager identity or force availability refreshes during database reload.

- [P2] Do not swallow DB-backed source failures — server/src/services/DbManager.js:608-608
  When a DB-backed search source throws because of a connection, schema, or query error, this broad `allSettled` path filters the rejection exactly like the intentionally unsupported pure-endpoint source and returns partial or empty results without a warning. The same pattern was added to `getOne` and submission cells; skip only known unbound endpoint sources or use the existing source runner so genuine SQL failures remain visible.

- [P2] Serialize endpoint quest rewards as the declared String — server/src/models/pokestopScanMapper.js:86-86
  When a GraphQL caller selects `quests.quest_rewards` for endpoint-backed Pokéstops, this field remains Golbat's parsed array, but the schema declares it as `String`; GraphQL therefore reports `String cannot represent value` and returns a field error, whereas SQL-backed results return the JSON string. Keep the native array for internal parsing but stringify the value assigned to the returned quest.

jfberry:
Six fixed, one is a repeat that's already handled.

### Fixed (`f8c29173`, `e95142be`)

- **Serialize endpoint quest rewards as String** — `quest_rewards` and `quest_conditions` are stringified in the scan mapper so the GraphQL `String` fields coerce (they're raw JSON-string columns on the SQL path); `parseRdmRewards` already handles a string internally.
- **Manual Pokéstop before capped candidates** — the off-viewport by-id fort is now **prepended** to the scan result, so `secondaryFilter`'s `resultLimit` cap examines it first and the deep link survives a dense viewport (gym/pokestop/station).
- **Invalidate availability TTL on manager change** — `loadLocalContexts` (startup/reload) now forces `setAvailable`, so a hot config reload with a fresh manager isn't short-circuited by the category-keyed TTL. The per-session stampede path stays non-forced/TTL-protected.
- **Preserve manager-owned metadata on empty refreshes** — `getAvailable` no longer overwrites quest conditions / rarity with a total-failure empty result (`runScannerSources` returns only fulfilled sources, so empty ⇒ every source failed → retain last-good).
- **Don't swallow DB-backed source failures** — `getOne` / `search` / `submissionCells` now route through `runScannerSources`, which isolates the (expected) unbound pure-endpoint rejection while **logging** genuine SQL-source failures, instead of the raw `allSettled` that hid them.
- **Keep rocket-reward DNF a true superset** — the `a<pokemon>` filter now narrows by **incident display type only** (`incident_display_type: [1,2,3,4]`, a safe rocket-incident superset), never by the reward pokemon. Display type is reliable (it comes from the GMO); the reward is a slot-1-only value from an optional invasion check, so a reward-derived `incident_character` could drop a stop `secondaryFilter` accepts (e.g. across a rotation). `secondaryFilter` confirms the actual reward from its slot/metadata data. This removes the earlier metadata-grunt expansion and its empty-map poison.

### Not changed

- **Scan limit before residual filters** — this is the same point from the prior two rounds. It's addressed by a dedicated, operator-tunable Golbat cap (`max_fort_results`, default 9000, independent of `max_pokemon_results`); the residual gates (area polygons, freshness, adv conditions) can't be pushed server-side, so the deliberate design is a high tunable ceiling rather than over-fetching/pagination, which would broaden every scan.

All lint/`node --check` clean; the DNF, mapper serialization, and availability paths are unit-checked. Assumes Golbat ≥ #385 + #382.

Mygod:
The endpoint station path bypasses the SQL field projection, while successful empty availability snapshots leave stale filters indefinitely. The new refresh TTL is also unavailable through the repository's environment-variable interface.

Full review comments:

- [P1] Strip station totals from non-dynamax results — server/src/models/Station.js:795-795
  When an endpoint-backed request uses the all/inactive station layer for a caller without `dynamax`, spreading the raw Golbat object retains `total_stationed_pokemon` and `total_stationed_gmax`; later processing clears the battle fields but not these totals, while the SQL path selects them only when `includeBattleData` is true. Because GraphQL callers choose the Station selection set, they can request these fields despite the permission projection. Construct the base projection explicitly and add the totals only under `includeBattleData`.

- [P2] Accept successful empty availability snapshots — server/src/services/EventManager.js:139-139
  When the last active option in a category disappears, such as when `stations.battles` becomes empty, `Db.getAvailable` legitimately returns `[]`, but this branch treats that exactly like an all-source failure. The previous persisted list remains visible and `availableUpdatedAt` is never set, so every session retries while clients keep receiving stale filters. Carry source-success status separately and clear the category on a successful empty snapshot.

- [P2] Expose the refresh TTL through environment config — config/default.json:62-62
  Container deployments using `custom-environment-variables.json` cannot override this new throttle because no mapping was added for `api.availableRefreshSeconds`, even though every other non-array scalar default has one. Consequently an `API_AVAILABLE_REFRESH_SECONDS` setting is ignored and the endpoint-scan throttle remains fixed at 60 seconds; add a numeric mapping alongside `queryUpdateHours`.

jfberry:
All three fixed in `2cd14701`.

- **Strip station totals from non-dynamax results** — the station endpoint spread the raw Golbat record, so `total_stationed_pokemon` / `total_stationed_gmax` rode along regardless of perms while the SQL path selects them only under `includeBattleData`. A new `finalizeEndpointStation` strips both when battle data isn't included, so a GraphQL selection can't read the stationed/gmax counts without the dynamax perm.
- **Accept successful empty availability snapshots** — good catch on the flip side of the earlier empty-refresh guard. `getAvailable` now distinguishes the two empty cases: it returns `null` on **total source failure** (every source rejected → caller keeps its last-good drawer) and `[]` on a **genuine empty snapshot** (sources succeeded, no active options → commit and clear). So a category whose last option disappears (e.g. `stations.battles` emptying) now clears instead of showing stale filters, while a transient outage still retains. `EventManager` retains only on `null` and the `/api/v1/available` route guards its sorts with `|| []`.
- **Expose the refresh TTL through env config** — added the `availableRefreshSeconds` → `API_AVAILABLE_REFRESH_SECONDS` (`number`) mapping in `custom-environment-variables.json`, alongside the other scalar `api` defaults, so `API_AVAILABLE_REFRESH_SECONDS` is honored.

Lint / `node --check` / JSON all clean.

Mygod:
Endpoint-backed scans can omit valid markers because they cap Golbat results before ReactMap's local gates, and explicit availability refreshes can reuse stale snapshots. The exported availability contract also diverges from the consumed Golbat payload.

Full review comments:

- [P2] Apply the scan limit after local gates — server/src/models/Pokestop.js:886-886
  When the viewport contains more than `queryLimits.pokestops` DNF matches—especially for an area-restricted user—Golbat stops its R-tree walk as soon as this request-side limit is reached. ReactMap applies `filterRTree`, freshness, and secondary filters afterward, so rejected stops consume the cap and valid stops later in the scan never arrive; the same pre-filter limit is sent in `Gym.js` and `Station.js`. Apply the final marker limit after these gates or move the gates into the endpoint.

- [P2] Bypass the snapshot cache on forced refreshes — server/src/utils/fortAvailable.js:50-50
  When `PUT /api/v1/available` or a configuration reload runs within 30 seconds of a previous fort refresh, `force=true` bypasses only `EventManager`'s TTL; this cache hit still returns the old resolved snapshot or cached failure without querying Golbat. The route can report success with stale options, and `API_AVAILABLE_REFRESH_SECONDS` values below 30 seconds cannot take effect for fort categories. Preserve batch coalescing while allowing forced refreshes to start a fresh cache generation.

- [P3] Align the availability types with Golbat's response — packages/types/lib/server.d.ts:84-90
  When typed code consumes the current `fort-scan-map-data` payload, this interface rejects it: invasion tuples include slot 2/3 fields and have no `count`, while lure and showcase tuples also omit `count`. The mapper already reads slot 2/3, so these exported declarations and the analogous mapper JSDoc contracts are inconsistent with the endpoint and prevent fixtures or consumers from being type-checked accurately.

jfberry:
All three fixed.

**[P2] Apply the scan limit after local gates** (`a8564fbd`) — the fort scans now send `limit: 0`, so Golbat traverses up to its own server-side `max_fort_results` backstop rather than stopping at the display limit before ReactMap's local gates run. `queryLimits.<type>` is now applied purely as a display cap *after* `filterRTree`/freshness/`secondaryFilter`:
- pokestop — already capped via `secondaryFilter`'s `resultLimit`
- gym — `final.slice(0, queryLimits.gyms)`
- station — `stations.slice(0, queryLimits.stations)`

This mirrors the SQL path, where `.limit(queryLimits.<type>)` runs after the `WHERE` clause, and relies on Golbat's `max_fort_results` as the true traversal bound.

**[P2] Bypass the snapshot cache on forced refreshes** (`be254651`) — `PUT /api/v1/available` and config reloads now call `bustCombinedFortCache()`, clearing the combined-snapshot map so the forced `setAvailable` batch starts a fresh generation instead of reusing a cached window. Batch coalescing is preserved (the concurrent fort batch still shares one Golbat request via the repopulated entry). The window now tracks `api.availableRefreshSeconds`, so sub-30s throttles take effect for fort categories.

**[P3] Align the availability types with Golbat's response** (`be254651`) — `AvailablePokestopInvasion` gains `slot2_*`/`slot3_*` and drops `count`; the lure and showcase tuples drop `count` too (quest keeps it). The mapper JSDoc typedefs were brought in line with the same shape.

Mygod:
Forced availability refreshes can retain data from a replaced source, and endpoint-backed station data loses or invents user-visible battle metadata. These are functional regressions despite the successful lint and build checks.

Full review comments:

- [P2] Do not reuse stale work for forced availability refreshes — server/src/services/EventManager.js:108-108
  When a database/endpoint hot reload or explicit PUT /api/v1/available overlaps a category refresh, this check runs before force is considered and returns the promise created with the old Db and cache generation. The forced call therefore never queries the newly installed manager, and the old result can be committed and TTL-stamped after reload; make pending work manager/generation-aware or let forced refreshes supersede it.

- [P2] Supply updated timestamps for endpoint station battles — server/src/models/Station.js:812-813
  For endpoint-backed upcoming or hidden battle cards, Golbat's ApiStationBattleResult has no updated field, so forwarding each battle unchanged through enrichStationBattle leaves battles[].updated undefined. The SQL path selects station_battle.updated, the GraphQL query requests it, and StationBattleTimer uses it for last_seen; populate the per-battle timestamp or extend the endpoint payload before mapping.

- [P2] Skip boss keys for battles without a known Pokémon — server/src/models/stationAvailableMapper.js:16-16
  When Golbat reports an active or upcoming battle before its boss is known, it encodes pokemon_id and form as 0, so this publishes 0-0 alongside the valid tier key. EventManager.addAvailable then treats that numeric key as a real Pokémon and injects id 0 into the masterfile/filter catalog, while endpoint battle records carry a null boss and matchesStationBattleFilter rejects them, making the advertised filter nonfunctional; only add the Pokémon key when pokemon_id > 0.

jfberry:
All three fixed in `e0e2dabf`.

**[P2] Forced availability refreshes no longer reuse superseded work** (`EventManager.js`) — the single-flight guard now applies only to non-forced calls; a forced refresh (hot reload / `PUT /api/v1/available` / scheduled interval) always starts a fresh refresh against the current `Db`, and a monotonic per-category generation token gates the commit so an older in-flight refresh can't commit or TTL-stamp its (old-manager) result after a reload. Non-forced session-init stampedes still coalesce on the shared pending promise.

**[P2] Endpoint station battles carry `updated`** (`Station.js`) — Golbat's scan battle has no per-battle `updated`, so it's now populated from the station's own `updated` (`b.updated ?? apiStation.updated ?? null`), mirroring the SQL `getFallbackStationBattle` path so `StationBattleTimer` has a `last_seen`.

**[P2] Skip boss keys for unknown bosses** (`stationAvailableMapper.js`) — `Number(b.pokemon_id) > 0` now guards the boss key, so an active/upcoming battle with an unknown boss (Golbat sends `null`, coerced to `NaN` here) no longer advertises a `null-null`/`0-0` filter option that no marker can match; the tier key (`j{level}`) is still published while the boss is unknown.

Mygod:
Thanks. The unknown-boss guard is functionally correct, but I don’t think the other two findings are fully resolved.

For `battles[].updated`, the fix belongs in Golbat. These are multi-battle records, whose SQL equivalent uses `station_battle.updated`; `station.updated` is only the legacy single-battle fallback. Copying one station-wide timestamp onto every battle fabricates `last_seen` values. Golbat already stores `StationBattleData.Updated`, so it should expose that in `ApiStationBattleResult` and ReactMap should pass it through unchanged.

The generation token fixes the immediate stale `Event.available` commit, but forced refreshes can now overlap on the same `Db`. `Db.getAvailable()` mutates `questConditions` and rarity before the generation check, allowing discarded work to overwrite current metadata. Scheduled intervals also capture the original `Db` and are not necessarily restarted on a database-only reload, so a later timer can still commit data from the replaced manager. The refresh lifecycle needs to be manager-aware, with same-manager work serialized or committed as one atomic snapshot.

The `pokemon_id > 0` guard can remain. Please correct its explanation/JSDoc: Golbat availability uses `0` as the unknown sentinel, `Number(null)` is `0` rather than `NaN`, and the payload has non-null `pokemon_id`/`form` fields with no `count`.

jfberry:
Thanks for your observation about leaking sentinals, that's a poor api - I'll fix that in golbat and push the reactmap changes later; other comments are bang on, they'll be fixed at the same time

Mygod:
Hello fellow human

jfberry:
Thanks — all addressed, and the `updated`/sentinel ones went into Golbat as you noted they should (companion PR `UnownHash/Golbat#385`, `959cb17`).

**`battles[].updated`** — fixed in Golbat: `ApiStationBattleResult` now exposes the stored per-battle `StationBattleData.Updated`, and ReactMap passes it through unchanged, so multi-battle cards get the real `station_battle.updated` rather than a station-wide fallback.

**Availability `0` sentinel** — rather than only correcting the guard's comment, Golbat's availability output now emits `null` where the DB column is `NULL` (unknown boss / egg / type-based showcase / unconfirmed slot), matching what a consumer would read from the row. The pokémon/form pair moves together (null id ⇒ null form; a present id keeps `form 0`). ReactMap's existing `> 0` guards already handle it; the JSDoc/types are now `number | null`, and yes — `Number(null)` is `0`, not `NaN`, and the battle availability struct has no `count`.

**Generation token** — extended to close both gaps you flagged:
- `getAvailable` is now pure and returns `{ available, conditions?, rarity? }`; EventManager commits the drawer **and** the manager metadata (`questConditions`/`rarity`) together under the generation gate, so a superseded refresh can no longer mutate metadata ahead of the check.
- A database-only reload now restarts the scheduled intervals, so a timer's captured `Db` is rebound to the freshly installed manager instead of committing the replaced one's data.

jfberry:
These versions deployed on my prod - no raids atm, but everything working as expected otherwise

Mygod:
The endpoint-backed paths introduce request amplification on failed searches, persist expected by-ID misses as errors, and disable healthy endpoints when a dual source's DB probe fails. These are operational and fallback correctness regressions.

Full review comments:

- [P1] Stop retrying after all search sources reject — server/src/services/DbManager.js:615-618
  When every source rejects—normally for a pure-endpoint Gym/Pokestop/Station source whose search method has no endpoint branch, or during a DB outage—`runScannerSources` collapses the batch to `[]`, so this loop treats the failure as a successful empty radius and keeps widening. With default limits, one Gym search can issue about 557 failing queries and warning logs before returning empty; preserve an all-failed signal or skip unsupported sources and stop expanding.

- [P2] Exclude fort-by-id 404s from failed-request logging — server/src/utils/evalScannerQuery.js:100-107
  When a manual or deep-linked fort ID is absent from an endpoint—especially with multiple scanner sources—Golbat returns 404 and this helper routes it through `fetchJson`, whose normal-miss exemption only covers `/api/pokemon/id`. The expected miss therefore logs an error and writes a timestamped request file under `logs/` before becoming `null`; repeated misses create unnecessary log noise and files, so fort-by-id 404s should be handled as normal misses.

- [P2] Initialize endpoint context before probing the fallback DB — server/src/services/DbManager.js:329-332
  When a dual endpoint+DB source's schema check rejects, such as during a transient scanner DB outage, control jumps to the outer catch before this overlay runs. The source is left without `mem`, `secret`, or `httpAuth`, causing endpoint-capable fort queries to bypass the healthy endpoint and execute against the failed DB; initialize endpoint context independently of the DB capability probe.

Mygod:
> These versions deployed on my prod - no raids atm, but everything working as expected otherwise

Unfortunately, the systems you touched have a fuckton of features + edge cases etc.

jfberry:
> > These versions deployed on my prod - no raids atm, but everything working as expected otherwise
> 
> Unfortunately, the systems you touched have a fuckton of features + edge cases etc.

Yes, and you didn't want me to build a test suite lol

jfberry:
Raids availability working correctly also. So lmk if any other review angles pick up anything

jfberry:
All three fixed in `b197123a` (ReactMap-only, no Golbat changes).

**[P1] Stop retrying after all search sources reject** (`DbManager.search`) — `runScannerSources` returns `[]` only when every source rejected (a genuine-empty source is fulfilled with `[]`, giving `[[]]`), so it's a clean all-failed signal: the loop now breaks on `data.length === 0` instead of treating it as an empty radius and widening. That kills the ~557 failing queries + warnings an all-pure-endpoint fort search (or a DB outage) used to spin before the 2s cap.

**[P2] Exclude fort-by-id 404s from failed-request logging** (`fetchJson`) — the normal-miss exemption was `/api/pokemon/id` only; it's now `/\/api\/(pokemon|gym|pokestop|station)\/id\//`, so a `getOne` / manual / deep-link miss on a fort id returns quietly (mirroring an empty SQL lookup) instead of logging an error and writing a timestamped request file under `logs/`.

**[P2] Initialize endpoint context before probing the fallback DB** (`getDbContext`) — `schemaCheck` now runs in its own `try/catch` (degrading the SQL capability flags to `{}` on failure), and the endpoint overlay (`mem`/`secret`/`httpAuth`) + source assignment run regardless. So a transient scanner-DB outage no longer strips a dual source's endpoint context — its fort queries keep using the healthy endpoint instead of falling back to the down DB.

Mygod:
Endpoint-backed fort scans can silently truncate valid markers, and the endpoint gym path breaks the no-badge filter. These are functional regressions in the newly added endpoint flows.

Full review comments:

- [P1] Do not treat limit zero as an unbounded scan — server/src/models/Pokestop.js:892-892
  When a viewport's DNF matches more than Golbat's configured max_fort_results (9000 by default), limit: 0 uses that finite server cap rather than disabling the cap. Golbat stops its R-tree traversal before the area, freshness, and secondaryFilter gates here, so rejected candidates can consume the cap and silently omit valid markers even when fewer than ReactMap's display limit survive; Gym and Station send the same value. Use pagination, an actual unbounded producer mode, upstream filtering, or truncation detection with fallback.

- [P2] Exclude badged gyms from the endpoint's none view — server/src/models/Gym.js:580-580
  When an endpoint-backed user with existing badges enables only the badge layer with badge = none, this candidate path does not reproduce the SQL query's whereNotIn(...userBadges). The shared secondaryFilter accepts every candidate through (actualBadge === 'none' && onlyGymBadges), so previously badged gyms are returned too. Exclude the userBadges IDs here or make the final predicate verify badge absence.

jfberry:
For the first point, the maximum limit is aligned with the model for pokemon settings as well - and is a very high limit for forts. The AI suggestion is stupid. If individual users actually have a real problem with this (ha!) they can increase the limit.
I'll look at the second point.

jfberry:
Fixed in `e7ea3d8e`.

**[P2] Exclude badged gyms from the endpoint's `none` view** (`Gym.js`) — the endpoint candidate set isn't pre-filtered by badge (per-user ReactMap data Golbat can't know), so the shared `secondaryFilter`'s badge clause `(actualBadge === 'none' && onlyGymBadges)` was always true and let previously-badged gyms through, unlike the SQL path's `whereNotIn(userBadges)`.

Replaced the always-true clause with one that verifies badge membership per view, mirroring the SQL `whereIn`/`whereNotIn`:

```js
onlyGymBadges && (actualBadge === 'none' ? !newGym.badge : !!newGym.badge)
```

So the `none` view keeps only gyms the user has **no** badge for, and a specific/`all` badge view keeps only those they **do** (`newGym.badge` is only ever set when `onlyGymBadges`). It's a no-op for the SQL path — those results are already `whereIn`/`whereNotIn`'d — and closes the endpoint leak. Verified equivalent across every badge view (none / tier / all / off) for both the SQL and endpoint paths.

Mygod (approved):

jfberry:
Golbat has had the requisite PR merged down to main

PR #1237, opened 2026-08-10
Fix Rocket reward filter identity and add quest task filtering

DannyM300:
## Summary

Hi! This started with a user report that some Rocket reward Pokémon appeared more than once in the filter menu and that only one of the identical-looking toggles affected the displayed invasion.

While investigating and testing that fix, I also added an optional task-first quest filter requested by users, plus a profile payload fix needed to save the additional dynamic filters reliably.

I have tried to keep the SQL and Golbat/API paths consistent throughout. I appreciate that the maintainers know this codebase better and may prefer a different implementation or may want the changes split differently.

### Root cause

ReactMap can build available Rocket reward filters from two sources:

- confirmed scanner incident data;
- the invasion masterfile fallback.

Those sources can describe the same reward Pokémon with different form IDs. For example, Taillow could be represented as both:

```text
a276-0
a276-3163
```

The menu label only uses Pokémon ID `276`, so both entries were displayed as **Taillow**. However, their internal keys remained different, and a JavaScript `Set` could not deduplicate the two strings.

Filtering then treated the same identity inconsistently:

- the SQL prefilter reduced Rocket reward keys to Pokémon species ID;
- the later in-memory invasion matcher required the complete species-and-form key.

Consequently, one duplicate could match confirmed scanner data while the other matched the masterfile/unconfirmed path. Both looked identical to the user, but one could appear to do nothing for the currently displayed grunt.

### Fix

This canonicalises and deduplicates Rocket reward filter keys across data sources:

- Rocket rewards are matched by Pokémon species ID, independent of scanner/masterfile form differences.
- A stable canonical form is taken from the masterfile where available, retaining one compatible display/storage key.
- Available Rocket reward keys are deduplicated by species on both the SQL and Golbat `/api/fort/available` paths.
- Hidden automatically generated form-specific sibling filters are removed, leaving the visible species toggle authoritative.
- Legacy form-less and form-specific saved filter keys remain accepted.
- The existing configured default for Rocket reward filters is preserved.

The intended matching remains additive: an invasion is shown when either its grunt type or one of its possible reward Pokémon is enabled, and hidden when neither is enabled.

A longer explanation and reproduction example are included in `docs/rocket-reward-filter-bug.md`.

## User-requested quest task filtering

This adds a **Tasks** tab alongside the existing Items and Pokémon quest reward filters. It allows users to select quests by task rather than only by reward.

Task filters use a stable `k<title>-<target>` key and support:

- enabling or disabling a complete task;
- narrowing a task to particular rewards through the advanced dialog;
- retaining the existing inverse behaviour where a reward can be narrowed to particular task conditions.

Task-to-reward mappings are generated for both SQL-backed available PokéStop data and Golbat `/api/fort/available` data. For the Golbat DNF path, enabled task keys are expanded into the relevant reward keys before query translation; the SQL/in-memory path applies the equivalent task-and-reward match directly.

Task filters default to **off**, making the feature opt-in and preserving the behaviour of existing reward filters.

## Profile payload/save fix

Available filters are dynamically merged into the client filter tree. Serialising the entire merged tree into every profile includes a large number of entries that are identical to the server defaults; adding task filters made this large enough to trigger an oversized database insert in testing.

Profile creation and updating now:

- compare the current filters with the server-provided defaults;
- store only values that differ from those defaults;
- retain unknown keys for forwards/backwards compatibility;
- rely on the existing default merge when loading a profile;
- report update failures consistently with create failures.

This keeps users' actual selections, including task selections, while avoiding a MySQL/MariaDB schema change.

## Validation

- Confirmed the branch merges cleanly with current upstream `main` (`8ce212ff`).
- Ran 33 focused automated tests covering Rocket key matching/canonicalisation/deduplication, task matching, advanced narrowing and Golbat DNF expansion.
- Built the combined branch against current upstream successfully with Vite (2,375 modules transformed).
- Manually tested SQL and Golbat-backed data, confirmed and unconfirmed invasions, task/reward filtering, and profile create/update/load behaviour on a test instance.

Happy to adjust the implementation or split the Rocket fix, task filtering and profile compaction if that would be easier to review.


Mygod:
Please ask your Claude to open each independent thing on its own PR.

DannyM300:
Thanks — understood. I'll split the Rocket fix, profile payload compaction, and task filtering into focused PRs so each can be reviewed independently.

DannyM300:
Split into focused draft PRs as requested:

- #1238 — Rocket reward filter identity
- #1239 — profile filter payload compaction
- #1240 — opt-in quest task filtering

PR #1238, opened 2026-08-10
Fix Rocket reward filter identity across scanner sources

DannyM300:
## Summary

Fix duplicate and ineffective Rocket reward Pokémon filters by giving each reward species one stable, source-independent filter identity.

This started from a user report where two entries with the same label could appear in the Rocket Pokémon menu, but only one appeared to affect the currently displayed invasion.

## Root cause

Available Rocket rewards are assembled from confirmed scanner incident data and the invasion masterfile fallback. Those sources can report the same Pokémon with different form IDs. For example, Taillow could produce both:

```text
a276-0
a276-3163
```

Both keys render as **Taillow** because the label uses only Pokémon ID `276`, but a `Set` cannot deduplicate the different complete strings.

Filtering then treated that identity inconsistently:

- the SQL prefilter reduced Rocket keys to Pokémon species ID;
- the secondary in-memory matcher required the complete species-and-form key.

As a result, one identical-looking toggle could match confirmed scanner data while the other matched the masterfile/unconfirmed path.

## Changes

- Match Rocket rewards by Pokémon species ID, independent of scanner/masterfile form differences.
- Select one canonical masterfile form for the stable visible/storage key where available.
- Deduplicate Rocket reward keys by species in both SQL and Golbat `/api/fort/available` results.
- Remove hidden form-specific sibling filters so the visible species toggle is authoritative.
- Continue accepting legacy form-less and form-specific saved keys.
- Preserve the configured default value for Rocket reward filters.

The existing additive behaviour remains: an invasion is shown when either its grunt type or a possible reward Pokémon is enabled.

## Validation

- Rebased independently on current upstream `main` (`8ce212ff`).
- 14 focused Rocket identity, canonicalisation and deduplication tests pass.
- The complete combined feature set was also production-built successfully against the same upstream revision.

I appreciate the maintainers may prefer a different form of canonicalisation and am happy to adjust the implementation.

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -540,7 +540,7 @@
       "pokemon": true,
       "invasions": false,
       "allInvasions": true,
-      "invasionPokemon": false,
+      "invasionPokemon": true,
       "baseMegaEnergyAmounts": [],
       "stardust": {
         "min": 100,
--- a/config/default.json
+++ b/config/default.json
@@ -540,7 +540,7 @@
       "pokemon": true,
       "invasions": false,
       "allInvasions": true,
-      "invasionPokemon": true,
+      "invasionPokemon": false,
       "baseMegaEnergyAmounts": [],
       "stardust": {
         "min": 100,
--- a/server/src/filters/builder/base.js
+++ b/server/src/filters/builder/base.js
@@ -115,7 +115,6 @@ function buildDefaultFilters(perms) {
               hasDualQuestLayer && perms.pokestops ? false : undefined,
             standard: new BaseFilter(),
             filter: {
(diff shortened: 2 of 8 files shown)


Mygod:
> Match Rocket rewards by Pokémon species ID, independent of scanner/masterfile form differences

This PR will not be merged as is since it deletes intentionally built features. Also I cannot reproduce this issue locally.

Mygod:
Fixed via https://github.com/WatWowMap/ReactMap/commit/5ff1a483227d658f4f829a7ffe7bba3e9caac370 (and others).

PR #1243, opened 2026-08-21
fix(geocoder): stop a provider mismatch from crashing the server

TurtIeSocks:
## What happens today

A webhook whose `nominatimUrl` points at Photon without `geocoderProvider: "photon"` takes the process down rather than failing the request. Reported from production after #1242 shipped.

The error is `TypeError: Cannot read properties of undefined (reading 'suburb')`.

## Why

node-geocoder decides how to read a response by its shape: an array is a result list, anything else is a single result. Photon answers with a GeoJSON object, so the entire `FeatureCollection` is handed to `_formatResult` as though it were one place.

node-geocoder 4.4.1 guards its own address lookup and returns undefined fields rather than throwing. The patch ReactMap layers on top of it did not, so `result.address.suburb` threw.

That throw never reached `geocoder()`'s catch. node-geocoder resolves through bluebird's `asCallback`, so a throw inside `_formatResult` surfaces as an uncaught exception and kills the process. The `try`/`catch` reads as though every failure ends in `return {}`, and this one could not be caught there at all. Confirmed directly: a `try`/`catch` around the `await` still lets the process die.

Worth noting for anyone reading the lockfile: `"node-geocoder": "^4.2.0"` resolves to 4.4.1, and 4.2.0's `_formatResult` is *unguarded*. The two versions fail differently, and 4.4.1 is what actually runs.

## Changes

The `_formatResult` patch optional-chains the address, so a response without one cannot throw.

`nominatimGeocoder` awaits its results and rejects when the raw body is a GeoJSON `FeatureCollection`, with a message naming the fix. Rejecting from an async function reaches the existing catch normally, so the operator gets a log line instead of a dead process.

`photonGeocoder` does the mirror check for a JSON array, so the opposite mismatch reports itself rather than returning an empty result set with no reason.

`results.raw` survives node-geocoder's wrapper and separates the two cleanly: an Array for Nominatim, `type: "FeatureCollection"` for Photon.

## What this does not do

It stops the crash. It does not make a misconfigured webhook geocode. A Photon backend still requires the opt-in:

```json
{ "webhooks": [{ "nominatimUrl": "http://127.0.0.1:2322", "geocoderProvider": "photon" }] }
```

Auto-switching on the response shape was considered and left out on purpose. It would let a request round-trip decide behaviour and would hide the misconfiguration permanently, rather than surfacing it once.

## Testing

Four new cases drive `geocoder()` over a real HTTP server rather than testing the mapping in isolation, which is the gap that let this ship: both mismatched pairs, and both matched pairs to show the checks do not reject valid responses.

- `yarn lint` passes
- `yarn build` passes
- `yarn prettier` passes
- `node --test server/test/geocoder.test.js` passes 26/26

Removing the optional chaining hangs the test runner instead of failing it, because the process dies mid-run. That is the same fatality seen in production, and it is worth knowing that this particular regression would show up in CI as a timeout rather than a clean failure.

`yarn test` also runs `server/test/rocketPokemonFiltering.test.js`, which fails with `No database selected for React Map Tables`. That is unrelated and predates this branch: `server/src/db/knexfile.cjs` calls `process.exit(9)` at import when no schema has `user` in its `useFor`, and the test reaches it through `services/state`. It has failed on every CI run since it landed.

### DIFF
--- a/packages/types/lib/config.d.ts
+++ b/packages/types/lib/config.d.ts
@@ -201,9 +201,10 @@ export interface Webhook {
    * handling. Defaults to `nominatim`.
    *
    * Optional when Poracle reports it. Poracle sends the backend behind its
-   * `providerURL` as `provider`, and ReactMap uses that when this is unset,
-   * the same way `nominatimUrl` falls back to Poracle's `providerURL`. Setting
-   * it here always wins.
+   * `providerURL` as `provider`, and ReactMap uses that when this is unset and
+   * `nominatimUrl` was itself inherited from Poracle. A locally configured URL
+   * keeps its local backend, since the two are one setting in two fields.
+   * Setting this explicitly always wins.
    */
   geocoderProvider?: 'nominatim' | 'photon'
   trialPeriodEligible?: boolean
--- a/packages/types/lib/config.d.ts
+++ b/packages/types/lib/config.d.ts
@@ -199,6 +199,11 @@ export interface Webhook {
    * Which geocoding backend `nominatimUrl` points at. Photon speaks GeoJSON
    * rather than Nominatim's JSON, so it needs its own request and response
    * handling. Defaults to `nominatim`.
+   *
+   * Optional when Poracle reports it. Poracle sends the backend behind its
+   * `providerURL` as `provider`, and ReactMap uses that when this is unset,
+   * the same way `nominatimUrl` falls back to Poracle's `providerURL`. Setting
+   * it here always wins.
    */
   geocoderProvider?: 'nominatim' | 'photon'
   trialPeriodEligible?: boolean
--- a/server/src/services/Poracle.js
+++ b/server/src/services/Poracle.js
@@ -55,12 +55,21 @@ const REMOTE_GEOCODERS = new Set(['nominatim', 'photon'])
  * deployment that gets its geocoder URL from Poracle no longer has to restate
  * the backend type in ReactMap's own config. Local config still wins, matching
  * how `providerURL` and `addressFormat` already behave.
- * @param {string} [remote] The `provider` field from Poracle's remote config
- * @param {string} [local] geocoderProvider from this webhook's own config
+ *
+ * The backend type is only inherited when the URL was inherited with it. A URL
+ * and the protocol used to talk to it are one setting in two fields, and
+ * splitting them across two sources is how a working Nominatim deployment ends
+ * up being addressed as Photon: keep a local `nominatimUrl`, omit
+ * `geocoderProvider`, and let a Photon-backed Poracle supply the type.
+ * @param {object} args
+ * @param {string} [args.remote] The `provider` field from Poracle's remote config
+ * @param {string} [args.local] geocoderProvider from this webhook's own config
+ * @param {boolean} args.usingRemoteURL Whether nominatimUrl also came from Poracle
  * @returns {'nominatim' | 'photon' | undefined}
  */
-function resolveGeocoderProvider(remote, local) {
+function resolveGeocoderProvider({ remote, local, usingRemoteURL }) {
   if (local) return /** @type {'nominatim' | 'photon'} */ (local)
+  if (!usingRemoteURL) return undefined
   return REMOTE_GEOCODERS.has(remote)
     ? /** @type {'nominatim' | 'photon'} */ (remote)
     : undefined
@@ -265,13 +274,17 @@ class PoracleAPI {
     if (addressFormat && !this.addressFormat) {
       this.addressFormat = addressFormat
     }
-    if (providerURL && !this.nominatimUrl) {
+    // Evaluated before the assignment below, because it is the assignment that
+    // decides whether the backend type may be inherited too.
+    const usingRemoteURL = !!providerURL && !this.nominatimUrl
+    if (usingRemoteURL) {
       this.nominatimUrl = providerURL
     }
-    this.geocoderProvider = resolveGeocoderProvider(
-      provider,
-      this.geocoderProvider,
-    )
+    this.geocoderProvider = resolveGeocoderProvider({
+      remote: provider,
+      local: this.geocoderProvider,
+      usingRemoteURL,
+    })
     this.leagues = [
       { name: 'great', cp: 1500, min: remoteConfig.pvpFilterGreatMinCP },
       { name: 'ultra', cp: 2500, min: remoteConfig.pvpFilterUltraMinCP },
--- a/server/src/services/Poracle.js
+++ b/server/src/services/Poracle.js
@@ -41,6 +41,31 @@ const SUBCATEGORIES = /** @type {const} */ ({
   gym: ['raid', 'egg', 'gym'],
 })
 
+/**
+ * Geocoding backends ReactMap can drive. Poracle also offers `google` and
+ * `none`, which have no equivalent here, so they are ignored rather than
+ * mapped onto something they are not.
+ */
+const REMOTE_GEOCODERS = new Set(['nominatim', 'photon'])
+
+/**
+ * Resolves which geocoding backend a webhook should use.
+ *
+ * Poracle reports the backend behind its `providerURL` as `provider`, so a
+ * deployment that gets its geocoder URL from Poracle no longer has to restate
+ * the backend type in ReactMap's own config. Local config still wins, matching
+ * how `providerURL` and `addressFormat` already behave.
+ * @param {string} [remote] The `provider` field from Poracle's remote config
+ * @param {string} [local] geocoderProvider from this webhook's own config
+ * @returns {'nominatim' | 'photon' | undefined}
+ */
+function resolveGeocoderProvider(remote, local) {
+  if (local) return /** @type {'nominatim' | 'photon'} */ (local)
+  return REMOTE_GEOCODERS.has(remote)
+    ? /** @type {'nominatim' | 'photon'} */ (remote)
+    : undefined
+}
+
 class PoracleAPI {
   /** @param {import("@rm/types").Config['webhooks'][number]} webhook */
   constructor(webhook) {
@@ -231,7 +256,11 @@ class PoracleAPI {
         `Poracle must be at least version 4.8.4, current version is ${this.version}`,
       )
     }
-    const { providerURL, addressFormat, ...rest } = remoteConfig
+    // `provider` is pulled out of remoteConfig even when it is not usable.
+    // this.provider is the *webhook* provider ('poracle'); letting Poracle's
+    // geocoding provider through the spread below would silently overwrite it
+    // with an unrelated meaning.
+    const { providerURL, provider, addressFormat, ...rest } = remoteConfig
     Object.assign(this, rest)
     if (addressFormat && !this.addressFormat) {
       this.addressFormat = addressFormat
@@ -239,6 +268,10 @@ class PoracleAPI {
     if (providerURL && !this.nominatimUrl) {
       this.nominatimUrl = providerURL
     }
+    this.geocoderProvider = resolveGeocoderProvider(
+      provider,
+      this.geocoderProvider,
+    )
     this.leagues = [
       { name: 'great', cp: 1500, min: remoteConfig.pvpFilterGreatMinCP },
       { name: 'ultra', cp: 2500, min: remoteConfig.pvpFilterUltraMinCP },
@@ -1257,4 +1290,4 @@ class PoracleAPI {
   }
 }
 
-module.exports = { PoracleAPI }
+module.exports = { PoracleAPI, resolveGeocoderProvider }
--- a/server/src/services/geocoder.js
+++ b/server/src/services/geocoder.js
@@ -85,6 +85,27 @@ function assertNominatimResponse(results, url) {
       `${url} answered as a Photon instance rather than a Nominatim one. Set "geocoderProvider": "photon" on this webhook, or point the URL at a Nominatim instance.`,
     )
   }
+
+  // Anything else object-shaped that carries no Nominatim result fields is an
+  // upstream error body, such as RFC 7807 problem JSON from a proxy in front
+  // of Nominatim. node-geocoder has already mapped it onto one blank entry by
+  // the time it gets here, and letting that through hands the caller a result
+  // with null coordinates -- Location.jsx pans the map to it. Rejecting keeps
+  // geocoder()'s fallback shape ([] or ''), and the message stays generic
+  // because this is a failing upstream, not a misconfigured provider.
+  // {"error": ...} bodies never reach this: node-geocoder converts those to a
+  // thrown Error itself. A genuine /reverse result always carries lat,
+  // display_name and address (all 42 captured fixtures do).
+  const isNominatimResult =
+    'lat' in raw ||
+    'display_name' in raw ||
+    'address' in raw ||
+    'place_id' in raw
+  if (!isNominatimResult) {
+    throw new Error(
+      `${url} answered with an error body instead of a geocoding result: ${JSON.stringify(raw).slice(0, 200)}`,
+    )
+  }
 }
 
 /**
--- a/server/src/services/geocoder.js
+++ b/server/src/services/geocoder.js
@@ -68,9 +68,15 @@ function assertNominatimResponse(results, url) {
   //
   // node-geocoder ignores the status and parses the body regardless, so both
   // arrive here as an object with no address and format into a blank result.
+  // The 404 body is matched on Javalin's own documentation URL rather than on
+  // the generic title/status pair, because that pair is RFC 7807's shape: a
+  // rate-limiting proxy in front of a healthy Nominatim answers
+  // {"title":"Too Many Requests","status":429}, and diagnosing that as "your
+  // provider is wrong" would turn a transient failure into advice to break a
+  // working configuration permanently.
   const isPhoton =
     raw.type === 'FeatureCollection' ||
-    (typeof raw.title === 'string' && typeof raw.status === 'number') ||
+    (typeof raw.type === 'string' && raw.type.includes('javalin.io')) ||
     (typeof raw.message === 'string' &&
       raw.message.includes('Unknown query parameter'))
 
@@ -104,7 +110,13 @@ async function nominatimGeocoder(url, search, isReverse) {
       // `neighborhood` and formatter() templates on `neighborhoods`, so the
       // value reached neither consumer. Carrying the alias is what makes it
       // visible without changing what node-geocoder itself produces.
-      neighborhood: formatted.neighbourhood || '',
+      //
+      // quarter is the fallback because Nominatim reports neighbourhood-level
+      // data under that key for many places: 1521 N Hoyne Ave carries
+      // quarter "Wicker Park" and no neighbourhood at all, and node-geocoder
+      // only reads address.neighbourhood, leaving the pair empty.
+      neighbourhood: formatted.neighbourhood || result.address?.quarter || '',
+      neighborhood: formatted.neighbourhood || result.address?.quarter || '',
     }
   })(stockGeocoder._geocoder._formatResult)
   // Awaited rather than returned so the shape check runs here. A throw inside
@@ -159,10 +171,13 @@ async function geocoder(nominatimUrl, search, reverse, format, provider) {
     return reverse
       ? formatter(format, results[0])
       : format
-        ? results.map((result) => ({
+        ? // The mapped fields ride along rather than being rebuilt, because the
+          // Geocoder GraphQL type resolves neighborhood, suburb and the rest
+          // straight off this object. Returning only the formatted triple made
+          // every other field null whenever an addressFormat was configured.
+          results.map((result) => ({
+            ...result,
             formatted: formatter(format, result),
-            latitude: result.latitude,
-            longitude: result.longitude,
           }))
         : results
   } catch (e) {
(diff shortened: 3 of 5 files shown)


Mygod:
The primary Photon-on-Nominatim crash is addressed, but the newly added symmetric mismatch handling remains incomplete for the application's reverse-geocoding path.

Review comment:

- [P2] Reject Nominatim reverse objects on the Photon path — server/src/services/photonGeocoder.js:263-263
  When `geocoderProvider: 'photon'` is paired with a Nominatim endpoint that returns JSON for gym reverse geocoding, Nominatim `/reverse` returns a single result object rather than the array used by `/search`; the resolver invokes this path at `server/src/graphql/resolvers.js:467`. `Array.isArray(response)` therefore misses the mismatch, `features` becomes `[]`, and `geocoder()` produces only a generic formatter error/`{}` instead of the actionable provider error. Detect the reverse-response shape and add a reverse regression.

TurtIeSocks:
Right, fixed in 73d14bd7. The check only recognised Nominatim's `/search` array and missed its `/reverse` object, which is the shape every gym lookup goes through.

`isNominatimResponse` now covers both. The test is deliberately positive, looking for fields a Nominatim result carries rather than treating anything without `features` as suspect: `fetchJson` returns a Response object on a failed request, and a no-match answers with `{"error":"Unable to geocode"}`. Neither is a provider mismatch, and an empty result is already the correct outcome for both, so a miss must not be reported as a misconfiguration.

Three tests: a Nominatim reverse object and a Nominatim search array both raising the provider error, and a no-match object still resolving to an empty list.

Worth flagging how the first version of those tests went, since it bears on the rest of this PR. I wrote them against `geocoder()` and they passed against the unfixed code. `geocoder()` catches everything and returns `{}`, so a mismatch and a plain miss are identical from the outside; the entire change is about which error reaches the log. They now assert against `photonGeocoder` directly, and I confirmed they fail without the fix (29/30) rather than assuming.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Mygod:
The new classifier cannot observe a standard Nominatim reverse response because the request does not select JSON. Consequently, the provider-mismatch diagnosis remains broken for the scenario the commit intends to fix.

Review comment:

- [P2] Request JSON before checking the reverse-response shape — server/src/services/photonGeocoder.js:290-290
  When `photonUrl` is an unmodified Nominatim base URL—the exact misconfiguration targeted here—the generated `/reverse` request supplies `lat`, `lon`, and `limit` but no output `format`; Nominatim defaults this endpoint to XML. `fetchJson()` therefore rejects while parsing the body before `isNominatimResponse()` can inspect this JSON shape, so the caller still receives only the generic geocoding error. Add `format=json` or inspect the raw content type/body, and make the regression server enforce that request contract.

TurtIeSocks:
Correct, fixed in 170258ce. I checked it against the real service rather than reasoning about it, and it is worse than reported: the forward path was broken too.

Verified against nominatim.openstreetmap.org:

- `/reverse?lat=..&lon=..&limit=1` with no format returns `text/xml` with HTTP 200. `fetchJson` cannot parse that and returns `undefined`, so `isNominatimResponse` never saw a body.
- `/api?q=Denver&limit=10` returns **404**. Photon serves `/api`; Nominatim has no such endpoint at any format, so a Nominatim search array could never have arrived on the forward path either. That half was equally unreachable and I had not noticed.

The reverse request now sends `format=json`, which Photon ignores. For the forward path, `fetchJson` returns the Response itself on any failed request, so that is now surfaced as an error naming the URL and status rather than an empty result. That also covers a genuine Photon instance returning 500, which previously resolved to no results and no reason given.

On the request contract you asked for: the regression server now behaves like a real Nominatim host rather than answering the same JSON on every path, which is precisely why neither check showed as broken. It answers 404 on `/api`, and XML on `/reverse` unless `format=json` is present. A separate test asserts the outgoing request actually carries the parameter, since that is the part the classifier depends on and the part I got wrong. Removing `format=json` fails two tests, removing the status check fails a third.

That is twice now on this PR that the tests passed against code that could not work, both times because the stub was more accommodating than the real thing. Worth weighing when you decide how much of this to trust.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

TurtIeSocks:
I finally ran this against a live Photon index instead of a stub, and it contradicted three things at once, including the fix I pushed two hours ago. Corrected in d3d93473.

**format=json is reverted, because it broke Photon.** Photon rejects any parameter outside its own allow list with a 400, so that commit would have taken down every reverse lookup on a correctly configured instance. The live answer:

```
{"message":"Unknown query parameter 'format'.  Allowed parameters are:
 [include, debug, dedupe, query_string_filter, lon, layer, limit, osm_tag,
  distance_sort, geometry, exclude, lang, radius, lat]"}
```

Your diagnosis was right and the suggested remedy is not available. The reverse mismatch is recognised from the unparseable body instead, which needs no extra parameter. `fetchJson` does not catch that itself: it does `return response.json()` inside its try block, so the rejection escapes its own catch and arrives here as a throw rather than as `undefined`, which is the opposite of what I claimed in the previous comment. There is now a test asserting the outgoing request carries only parameters Photon accepts, so this cannot regress quietly.

**Two mapping errors surfaced in the same run.**

`locality` is Nominatim's `neighbourhood`, not a smaller settlement. The live index returns `locality: "Wicker Park"` alongside a city, and Nominatim's own answer for that building carries `quarter: "Wicker Park"`. A genuine hamlet does not arrive that way at all: Bootjack comes back as `osm_value=locality` with the name in `name` and no `locality` field. Treating it as a city fallback, which is what the earlier review asked for and I implemented without checking, would name a neighbourhood as a town.

`district` is Nominatim's `suburb`. For 1521 North Hoyne Avenue, Photon sends `district: "West Town"` and Nominatim sends `suburb: "West Town"` — the same string for the same building. `address.suburb` was left empty on my claim that the two were different OSM concepts. That claim was wrong, and a decision was made on it. Photon's `Manhattan` for the Statue of Liberty matches Nominatim's `suburb` there too.

Both now sit between the street and the settlement in the formatted address, where Nominatim puts the levels they correspond to. Nine live queries, forward and reverse, now produce output matching the captured Nominatim structurally, including `neighbourhood=Golden Triangle` and `suburb=Manhattan` exactly.

Every fixture in the tests is now a captured live response rather than one I constructed.

For the record on this PR: three of your findings were things my tests could not see, and in each case the reason was the same. I was validating against my model of Photon rather than against Photon, and the stub agreed with the model. The endpoint had been available to me for most of that time.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Mygod:
The patch turns transient Photon HTTP failures into an invalid GraphQL list result and leaves direct district searches without the newly supported suburb value.

Full review comments:

- [P2] Preserve a list-shaped fallback for Photon HTTP failures — server/src/services/photonGeocoder.js:321-324
  When a correctly configured Photon returns a transient non-2xx response such as 429 or 500, this newly throws. `geocoder()` catches it and returns `{}`, but `Query.geocoder` is declared `[Geocoder]`; GraphQL therefore emits `Expected Iterable` and nulls the field instead of preserving the previous `[]` result. Keep the status observable while retaining a list-shaped forward fallback, or deliberately propagate the original error.

- [P2] Promote direct district results before reading suburb — server/src/services/photonGeocoder.js:245-245
  When Photon returns a forward result whose own `properties.type` is `district`, such as a suburb search, its label is in `name` and the containing-hierarchy `district` field is absent. `resolveSelfReference()` still has no `district` entry, so this new read produces `suburb: ''`; `{{suburb}}` formats and the GraphQL `suburb` field lose the selected place. Promote direct district results and cover this response shape.

TurtIeSocks:
Both correct, fixed in 0e168b6f, and both verified against the live index rather than reasoned about.

**The list-shaped fallback.** You are right that surfacing a non-2xx as a throw made things worse for the common case: a transient Photon 429 or 500 went from an empty list to a discarded field. The fallback now matches the shape the caller's schema declares, `[]` for the forward path and `''` for reverse, since `Query.geocoder` is `[Geocoder]` and `Gym.formatted` is `String`. The failure stays observable, logged immediately before the fallback returns.

Worth separating what I broke from what I found: the `{}` fallback predates this branch and was already wrong for both callers, but it was only reachable when `nominatimUrl` was missing entirely. Making upstream failures throw is what turned a rare bug into a routine one, so the regression is mine even though the defect was not.

**The district promotion.** Confirmed live, `q=West Town Chicago`:

```json
{"name": "West Town", "type": "district", "osm_key": "place",
 "osm_value": "suburb", "city": "West Chicago Township", "county": "Cook County"}
```

The label is in `name` and there is no `district` field, so the value being searched for was the one value missing from the answer. `district` is now in the type map, and that query returns `suburb: "West Town"` rather than `''`. A `Wicker Park Chicago` search now fills both levels, `neighbourhood: "Wicker Park"` and `suburb: "West Town"`.

Both fail without their fix, 34/35 for the district entry and 32/35 with the object fallback restored.

This is the second finding on this PR that came from the `district` to `suburb` mapping I added yesterday, which I would not have added at all without your earlier review, and would not have got right without checking a live index. The pattern across this PR is consistent enough to be worth stating plainly: nearly every defect here came from me modelling Photon's behaviour instead of observing it.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Mygod:
The crash guard and failure-shape changes are useful, but standard provider-mismatch handling still misses Photon's real request contract. The new locality mapping also does not cross the public schema and formatter boundary.

Full review comments:

- [P2] Detect standard Photon errors on Nominatim routes — server/src/services/geocoder.js:41-41
  When a standard Photon base URL is accidentally left on the default Nominatim provider, node-geocoder calls `/search` for forward requests (Photon serves `/api`) and appends Nominatim-only parameters such as `format` and `addressdetails` on `/reverse`; Photon therefore returns 404/400 instead of a `FeatureCollection`. This `raw.type` check never fires, and a reverse `{message: ...}` error body is formatted into a blank result without the actionable mismatch warning. Detect the actual route/status/error shape and make the regression fixture enforce Photon's routes and parameter allow-list.

- [P2] Expose locality through the public neighborhood contract — server/src/services/photonGeocoder.js:249-249
  When Photon supplies `properties.locality`, this writes it only as `neighbourhood`, while the GraphQL contract exposes `neighborhood` (`server/src/graphql/typeDefs/map.graphql:25`) and `formatter()` recognizes the different key `neighborhoods` (`server/src/services/geocoder.js:15`). Consequently the newly mapped Wicker Park value is unavailable both to raw GraphQL clients and to configured address formats; add the consumer-facing alias/token and cover it through the public geocoder path.

TurtIeSocks:
Both correct, fixed in c1e868e5. I reproduced the first one against a live deployment before fixing it, and it is the exact failure that was reported from production.

**Photon errors on Nominatim routes.** You are right that the `raw.type` check could never fire. node-geocoder calls `/search`, which Photon does not serve, and `_forceParams` puts `format` and `addressdetails` on every request, which Photon rejects outright. From a live instance:

```
GET /search?q=..&format=json&addressdetails=1
  404 {"title":"Endpoint GET /search not found","status":404,
       "type":"https://javalin.io/documentation#endpointnotfound"}

GET /reverse?lat=..&lon=..&format=json&addressdetails=1
  400 {"message":"Unknown query parameter 'format'.  Allowed parameters are:
       [include, debug, dedupe, query_string_filter, lon, layer, limit,
        osm_tag, distance_sort, geometry, exclude, lang, radius, lat]"}
```

The Javalin body even carries a `type`, so the old check saw a field with the wrong value rather than no field. Both shapes are now recognised. The end-to-end reproduction that previously returned `[{"formatted":""}]` with nothing in the log now logs the URL and the setting that fixes it.

The regression server enforces Photon's contract as you asked, rather than accepting anything: `/api` and `/reverse` only, Javalin's 404 body elsewhere, and a 400 for any parameter outside the allow list. That stub is what makes these tests fail without the fix, 36/38.

**The neighborhood contract.** Also right, and worse than described: there are three spellings in play. node-geocoder emits `neighbourhood`, `Geocoder.neighborhood` is American, and `formatter()` templates on `neighborhoods`. Both providers now carry the alias and the formatter accepts either spelling, so the value reaches raw GraphQL clients and configured address formats. That also revives the field for Nominatim deployments, where it had been dead for the same reason long before this branch.

One thing your review surfaced indirectly: the parity test duplicated the `_formatResult` patch inline, so adding the alias to the real one made the copy drift. It is updated, with a note that the duplication is the point.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Mygod:
The patch can misdiagnose generic upstream failures and still loses neighborhood data for quarter-shaped Nominatim responses and formatted GraphQL calls. These are observable contract and diagnostic defects.

Full review comments:

- [P2] Narrow the Photon error signature — server/src/services/geocoder.js:73-73
  When a valid Nominatim service or its proxy returns an RFC 7807-style failure such as `{"title":"Too Many Requests","status":429}`, node-geocoder places it in `results.raw`, and this generic predicate reports the host as Photon. The log then advises changing `geocoderProvider`, potentially turning a transient error into a persistent misconfiguration; match Photon-specific Javalin fields or route text instead.

- [P2] Preserve quarter-level Nominatim neighborhoods — server/src/services/geocoder.js:107-107
  When Nominatim returns neighborhood-level data as `address.quarter`—as in the Wicker Park response documented in `photonGeocoder.js:24-27`—node-geocoder leaves `formatted.neighbourhood` empty because it only reads `address.neighbourhood`. This alias therefore still publishes an empty `neighborhood`; include `result.address.quarter` as the fallback.

- [P2] Keep the alias on formatted GraphQL results — server/src/services/photonGeocoder.js:254-254
  When `webhook.addressFormat` is non-empty, the GraphQL resolver passes it to `geocoder()`, whose formatted branch returns only `formatted`, `latitude`, and `longitude` (`geocoder.js:162-166`). This newly added alias is therefore discarded before resolving `Geocoder.neighborhood`, leaving the field null despite the direct mapper test; preserve the mapped fields when adding `formatted` and exercise the resolver path as required by `AGENTS.md:9`.

TurtIeSocks:
All three correct, fixed in 70eedcd2.

**The narrowed signature.** The title/status pair is RFC 7807's shape, not Photon's, and your failure case is the damaging one: a rate-limiting proxy in front of a healthy Nominatim would have been diagnosed with advice to change `geocoderProvider`, converting a transient 429 into a permanently broken config. The 404 match now requires Javalin's documentation URL in the body's `type` field (`https://javalin.io/documentation#endpointnotfound`), which a real Photon sends and nothing else plausibly does. A 7807 body now resolves through the generic path as before, and a test pins each side: the proxy body must not throw the mismatch advice, the Javalin body still must.

**The quarter fallback.** Both spellings now fall back to `address.quarter`, so the Wicker Park response that motivated the alias actually populates it. You were right that the alias as shipped published an empty string for exactly the quarter-shaped responses it existed to surface.

**The formatted branch.** The result now spreads through with `formatted` added instead of being rebuilt as a triple, so `Geocoder.neighborhood`, `suburb`, and the rest resolve on the path every webhook user with an `addressFormat` takes. The resolver-path test exercises a formatted call end to end over a stub server and asserts `neighborhood`, `suburb`, and `streetNumber` survive alongside `formatted` — the direct-mapper test you called out was indeed testing a path the resolver never uses.

Each fix fails its test when reverted individually: 44/45, 44/45, and 42/45.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Mygod:
The narrowed classifier now treats transient problem JSON as a successful blank location. This breaks the public geocoder contract under the proxy-error scenario the commit explicitly targets.

Review comment:

- [P2] Reject non-Nominatim problem bodies after narrowing — server/src/services/geocoder.js:77-81
  When a Nominatim proxy returns the RFC 7807 JSON this change targets, such as a 429 body, node-geocoder maps the object to one blank entry before this check. Falling through now makes `geocoder(..., false, ...)` return a GraphQL option with null coordinates rather than `[]`; `Location.jsx:198-214` can then pass `[null, null]` to `map.panTo`. Keep the Photon-specific diagnostic narrow, but reject other non-Nominatim problem bodies with a generic upstream error.

TurtIeSocks:
Correct, fixed in 6b37fd32. Narrowing reopened the exact path the broad match had been closing, badly: the 7807 body fell through, node-geocoder had already mapped it onto one entry with undefined coordinates, and `geocoder()` resolved with that instead of `[]` — straight into `Location.jsx` panning to `[null, null]`.

An object body carrying none of `lat`, `display_name`, `address` or `place_id` now rejects with a generic upstream error that names the body. Generic on purpose, per your previous finding: a failing upstream is not a misconfigured provider, so the message carries no `geocoderProvider` advice. `{"error": ...}` bodies never reach the check — node-geocoder converts those to a thrown Error itself — and all 42 captured Nominatim fixtures carry `lat`, `display_name` and `address`, so genuine reverse results pass.

The test now asserts three things: the rejection fires, its message is not the provider advice, and the public entry point returns `[]` rather than a result with null coordinates. Fails 44/45 without the guard.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Mygod:
Will review the new commit in a bit. This is for the old.

The patch improves failure shapes but still turns Nominatim problem responses into invalid successful results and mislabels transient Photon failures as provider misconfiguration. Both issues affect normal operational failure paths.

Full review comments:

- [P2] Reject Nominatim problem bodies before formatting — server/src/services/geocoder.js:77-83
  When a valid Nominatim deployment is rate-limited with a JSON problem response such as `{"title":"Too Many Requests","status":429}`, node-geocoder ignores the HTTP status and this non-Photon path lets the object through. `geocoder()` consequently returns one blank `Geocoder` with null coordinates rather than `[]`, allowing the autocomplete to offer an unusable location. Treat non-Nominatim problem bodies as generic upstream failures while reserving provider-specific advice for Photon signatures, and exercise this through the public GraphQL path as required by `AGENTS.md:9`.

- [P2] Keep transient Photon failures out of provider diagnostics — server/src/services/photonGeocoder.js:331-334
  When a correctly configured Photon instance returns 429/500, or an authentication proxy returns 401/403, `fetchJson()` supplies a `Response` and this unconditional branch advises removing `geocoderProvider: "photon"`. Although the caller retains the correct fallback shape, the log misdiagnoses an operational failure as bad configuration and can prompt operators to break a working setup; reserve that advice for provider-specific evidence such as the Nominatim `/api` 404 and report other statuses as ordinary upstream failures.

Mygod:
The tested problem body is rejected, but the presence-only classifier still accepts malformed or extended problem bodies that produce the same null-coordinate result. The intended failure-path fix is therefore incomplete.

Review comment:

- [P2] Validate field values before accepting Nominatim bodies — server/src/services/geocoder.js:99-103
  When a problem response includes any one of these generic keys—RFC 7807 permits extension members, for example `{"title":"Too Many Requests","status":429,"address":null}`—this OR check accepts it as Nominatim. node-geocoder then formats it into an entry with undefined coordinates, so `geocoder(..., false, ...)` again exposes null GraphQL coordinates and `Location.jsx` can pan to `[null, null]`. Validate a complete, usable coordinate pair rather than key presence.

TurtIeSocks:
Both remaining findings fixed in fb398d78. (The first item of the earlier comment was the fall-through that 6b37fd32 addressed; the two below are the ones still open.)

**Coordinate validation over key presence.** You are right that the presence check was defeatable by design: RFC 7807 extension members mean `{"title":"Too Many Requests","status":429,"address":null}` walked straight through and reopened the null-coordinate result. The classifier now accepts a body only when `lat` and `lon` both parse to finite numbers — which all 42 captured fixtures do, and which is also the minimum a result must carry to be usable by any caller regardless of what else is in it. The extended-body case is pinned by a test through both `nominatimGeocoder` (generic rejection, no provider advice) and the public `geocoder()` path (`[]`), and a matched-pair test pins that a genuine reverse result still passes, so the stricter check cannot silently reject what it protects.

**Transient Photon failures.** Provider advice is now reserved for provider evidence: only the 404 on `/api` says anything about what the host is. A 429/500 from a struggling Photon or a 401/403 from an authentication proxy reports as `answered <status>; the upstream geocoder is failing, not misconfigured` — no `geocoderProvider` mention. Tested across all four statuses, asserting the advice is absent and the status is named; the 404 path keeps its existing test.

Non-vacuity confirmed for both: restoring the presence check fails one test (52/53), restoring the advice on non-404 statuses fails another (52/53).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Mygod:
The patch introduces a public provider-field asymmetry and false-positive provider diagnostics for valid or operational Photon responses. These can return inaccurate API data and prompt operators to break a valid backend configuration.

Full review comments:

- [P2] Stop treating every 404 as a Photon mismatch — server/src/services/photonGeocoder.js:338-340
  `photonGeocoder` enters this branch for both `/api` and `/reverse`, so a reverse lookup receiving 404 is incorrectly reported as a failure from `/api`. Moreover, Photon's documented [reverse-only mode](https://github.com/komoot/photon/blob/master/docs/usage.md#reverse-only-mode) intentionally returns 404 for `/api` while serving `/reverse`, making the “not a Photon instance” conclusion and removal advice wrong for a valid deployment. Report an endpoint/capability failure unless there is positive evidence of a provider mismatch.

- [P2] Keep generic JSON parse failures out of provider advice — server/src/services/photonGeocoder.js:317-319
  When `isReverse` is false, this catch is handling a parse failure from `/api`, yet it always explains Nominatim's `/reverse` XML behavior and recommends removing `geocoderProvider`. A correctly configured Photon endpoint or proxy returning a transient 200 HTML or truncated body follows the same path, turning an operational parse failure into harmful configuration advice. Restrict the mismatch diagnosis to identified reverse XML and keep other parse failures generic.

- [P2] Normalize provider before spreading formatted results — server/src/services/geocoder.js:204-206
  When `addressFormat` is truthy, this new spread exposes node-geocoder's synthetic `provider: 'openstreetmap'` through `Geocoder.provider`, while `formatPhotonFeature()` never supplies a provider, so equivalent Photon results resolve that public field as null. The parity test calls `_formatResult` directly and misses the later `_format()` step that adds this key. Add a Photon provider value or omit the field consistently before spreading.

PR #1244, opened 2026-08-23
feat(geocoder): derive geocoderProvider from Poracle's remote config

TurtIeSocks:
Stacked on #1243. Base is `fix/geocoder-photon-misconfig` so the shared test file does not conflict; retarget to `main` once that merges.

Requires the PoracleNG side, which adds `provider` to the `/api/config/poracleWeb` payload.

## The asymmetry this removes

`nominatimUrl` already falls back to Poracle's `providerURL`, so a deployment can take its geocoder URL from Poracle and never mention it in ReactMap's config. The backend *type* had no such fallback:

```js
if (providerURL && !this.nominatimUrl) this.nominatimUrl = providerURL   // derived
// geocoderProvider — local config only
```

Half the pair auto-populated and the other half did not. An operator whose URL came from Poracle still had to hand-set `geocoderProvider` locally, and if they did not, every request silently took the Nominatim branch against a Photon URL. That is not a hypothetical: it is what took a production instance down after #1242 shipped.

PoracleNG now reports the backend behind `providerURL` as `provider`, so `geocoderProvider` can follow the same rule as its URL. Local config still wins, matching how `providerURL` and `addressFormat` already behave.

## Ignored backends

Poracle also offers `google` and `none`. Neither has a ReactMap equivalent, so they resolve to `undefined` rather than being mapped onto `nominatim`, which would claim something untrue about the backend. The match is exact, so a differently-cased value is ignored rather than guessed at.

## The name collision

`provider` is destructured out of `remoteConfig` whether or not it is usable:

```js
const { providerURL, provider, addressFormat, ...rest } = remoteConfig
Object.assign(this, rest)
```

`this.provider` is the *webhook* provider, set from `webhook.provider` in the constructor. It is an unrelated field that happens to share a name with Poracle's geocoding provider, and letting the remote value through the spread would silently overwrite it. It is currently read nowhere else in ReactMap, so this would have been invisible until something started reading it.

## Testing

The resolution lives in a small exported function because `#fetchConfig` is private and would otherwise need a stubbed HTTP round trip to reach.

Four cases: the derivation, local precedence over the remote value, the ignored backends, and the webhook provider surviving a remote payload that carries a geocoding provider.

- `yarn lint` passes
- `yarn build` passes
- `yarn prettier` passes
- `node --test server/test/geocoder.test.js` passes 30/30

Not exercised against a live Poracle. The payload shape is taken from PoracleNG's `poracleWebResponse` struct and its OpenAPI golden, not from a real response, so the field name is worth confirming against a running instance before merge.

`yarn test` also runs `server/test/rocketPokemonFiltering.test.js`, which fails with `No database selected for React Map Tables`. Unrelated and predates this branch: `server/src/db/knexfile.cjs` calls `process.exit(9)` at import when no schema has `user` in its `useFor`.

### DIFF
--- a/packages/types/lib/config.d.ts
+++ b/packages/types/lib/config.d.ts
@@ -201,9 +201,10 @@ export interface Webhook {
    * handling. Defaults to `nominatim`.
    *
    * Optional when Poracle reports it. Poracle sends the backend behind its
-   * `providerURL` as `provider`, and ReactMap uses that when this is unset,
-   * the same way `nominatimUrl` falls back to Poracle's `providerURL`. Setting
-   * it here always wins.
+   * `providerURL` as `provider`, and ReactMap uses that when this is unset and
+   * `nominatimUrl` was itself inherited from Poracle. A locally configured URL
+   * keeps its local backend, since the two are one setting in two fields.
+   * Setting this explicitly always wins.
    */
   geocoderProvider?: 'nominatim' | 'photon'
   trialPeriodEligible?: boolean
--- a/packages/types/lib/config.d.ts
+++ b/packages/types/lib/config.d.ts
@@ -199,6 +199,11 @@ export interface Webhook {
    * Which geocoding backend `nominatimUrl` points at. Photon speaks GeoJSON
    * rather than Nominatim's JSON, so it needs its own request and response
    * handling. Defaults to `nominatim`.
+   *
+   * Optional when Poracle reports it. Poracle sends the backend behind its
+   * `providerURL` as `provider`, and ReactMap uses that when this is unset,
+   * the same way `nominatimUrl` falls back to Poracle's `providerURL`. Setting
+   * it here always wins.
    */
   geocoderProvider?: 'nominatim' | 'photon'
   trialPeriodEligible?: boolean
--- a/server/src/services/Poracle.js
+++ b/server/src/services/Poracle.js
@@ -55,12 +55,21 @@ const REMOTE_GEOCODERS = new Set(['nominatim', 'photon'])
  * deployment that gets its geocoder URL from Poracle no longer has to restate
  * the backend type in ReactMap's own config. Local config still wins, matching
  * how `providerURL` and `addressFormat` already behave.
- * @param {string} [remote] The `provider` field from Poracle's remote config
- * @param {string} [local] geocoderProvider from this webhook's own config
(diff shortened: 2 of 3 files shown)


Mygod:
The new fallback can pair Poracle's provider type with a different locally configured URL, regressing valid existing Nominatim configurations when Poracle uses Photon.

Review comment:

- [P1] Only inherit the provider with its URL — server/src/services/Poracle.js:271-274
  When `nominatimUrl` is set locally and `geocoderProvider` is omitted—the existing Nominatim-default configuration—while Poracle reports Photon, the preceding block keeps the local URL but this unconditional fallback installs `photon`. `geocoder()` then uses the Photon protocol against the local Nominatim endpoint, causing searches and reverse lookups to return `{}`. Only derive the remote provider when the remote `providerURL` was also selected, unless a provider was explicitly configured locally.

TurtIeSocks:
Correct, and this was a regression I introduced rather than a gap. Fixed in 98298a09.

The fallback was unconditional, so a webhook with a local `nominatimUrl` and no `geocoderProvider` inherited Photon from a Photon-backed Poracle while keeping its own Nominatim URL. That is the default shape of every existing Nominatim deployment, and it would have broken endpoints that were working fine before this PR.

The backend type is now inherited only when the URL was inherited with it. A URL and the protocol used to talk to it are one setting in two fields, and splitting them across two sources cannot produce a working pair except by luck. An explicit local `geocoderProvider` still wins in either case, which remains the only way to run a backend Poracle does not report.

`resolveGeocoderProvider` now takes a named argument object rather than a third positional boolean, so the pairing rule reads at the call site.

Three cases cover it: inheriting when the URL came from Poracle, refusing to inherit when it did not, and local config winning regardless. Confirmed failing without the guard (34/35) rather than assumed.

The irony is not lost on me. This PR exists to fix an asymmetry in how the pair is populated, and it shipped by breaking the pairing in the other direction.

The branch is rebased on the updated #1243, which now also carries the reverse-shape fix from your other review.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Mygod (approved):
The revised logic keeps the inherited provider paired with Poracle's URL while preserving explicit local configuration. Focused tests, lint, formatting, and the production build passed with no actionable regression found.

